PR #11, opened 2019-03-22
Use coroutines for AlertMover and MonitorRunner

vinooamzn:
This change converts the MonitorRunner and AlertMover to use kotlin coroutines. 

* MonitorRunner's blocking implementation and large thread pool have been removed without having to switch to a full callback style of API.  We no longer allocate & block a thread for the entire duration we run a monitor. We no longer block a thread when performing I/O on the ES cluster (performing the search or reading/writing alerts). The only blocking I/O that remains is when publishing notifications which is done on a much smaller background I/O threadpool. 
* AlertMover's callback based implementation is simplified considerably by coroutines. 

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -15,10 +15,17 @@
 
 package com.amazon.opendistroforelasticsearch.alerting
 
-import com.amazon.opendistroforelasticsearch.alerting.core.JobRunner
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertError
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertIndices
-import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertMover
+import com.amazon.opendistroforelasticsearch.alerting.alerts.moveAlerts
+import com.amazon.opendistroforelasticsearch.alerting.core.JobRunner
+import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
+import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob.Companion.SCHEDULED_JOBS_INDEX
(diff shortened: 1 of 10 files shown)


dbbaughe:
@vinooamzn Hi Vinoo, can you resolve the conflicts in this.

@jinsoor-amzn I believe you ran performance tests before for alerting to see the impact on a cluster, do you think it would be possible to run similar performance tests comparing alerting w/ threads and alerting w/ coroutines to confirm there is no negative impact for switching.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -93,74 +98,77 @@ class MonitorRunner(
     private val xContentRegistry: NamedXContentRegistry,
     private val alertIndices: AlertIndices,
     clusterService: ClusterService
-) : JobRunner {
+) : JobRunner, CoroutineScope, AbstractLifecycleComponent(settings) {
 
-    private val logger = ElasticAPI.INSTANCE.getLogger(MonitorRunner::class.java, settings)
+    private lateinit var job: Job
```
dbbaughe: As discussed offline: this job and the scheduled job (in runJob) both being referred to as job is confusing. Let's be more explicit with naming.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -93,74 +98,77 @@ class MonitorRunner(
     private val xContentRegistry: NamedXContentRegistry,
     private val alertIndices: AlertIndices,
     clusterService: ClusterService
-) : JobRunner {
+) : JobRunner, CoroutineScope, AbstractLifecycleComponent(settings) {
```
dbbaughe: Seems AbstractLifecycleComponent(settings) was [deprecated in 6.7 and completely removed in 7.0](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#_deprecated_constructor_literal_abstractlifecyclecomponent_settings_settings_literal_removed), they recommend using the parameterless constructor.

PR #17, opened 2019-04-01
Update LastFullSweepTime if the index doesn't exist

jinsoor-amzn:
*Description of changes:*
LastFullSweepTime does not get updated if the ScheduledJob Index doesn't exist. This commit is to fix that issue.


dbbaughe:
It looks like we are setting lastFullSweepTimeNano twice now in sweepAllShards (if index doesn't exist and at end of method). Can we just do it once at the start or is there some reason we should have it specifically once sweeping is done at the end?

jinsoor-amzn:
`lastFullSweepTimeNano` is at the bottom because if there is any exception that causes the Background Sweep to fail then `lastFullSweepTimeNano` wouldn't be updated. This way we will see the `lastFullSweepTimeNano` increasing indicating there is something going wrong with Backgound Sweep.

In case for index doesn't exists, the method `return` early so it doesn't update the `lastFullSweepTimeNano`. So this is to fix that issue where we don't skip updating `lastFullSweepTimeNano`.

dbbaughe:
Got it, LGTM!

Issue #22, opened 2019-04-04
No proxy support 


imranh2:
Notifications don't work when behind a proxy.

imranh2:
Think it might be as simple as sticking a .useSystemProperties() in the right when building the http client place to pull in http.proxyHost http.proxyPort...

imranh2:
Fixed in https://github.com/imranh2/alerting/commit/96ac0fb92cc630ffd16a4bbe02ca1cb45cf9a9a1

Will make a PR now.

imranh2:
PR was merged 🙂

PR #23, opened 2019-04-04
Add http proxy support to outgoing notifcations

imranh2:
Add proxy support for outgoing notifications by looking at system properties.

*Issue:* https://github.com/opendistro-for-elasticsearch/alerting/issues/22

*Description of changes:* Simply call useSystemProperties() which then finds and uses properties such as http.proxyHost, http.proxyPort, https.proxyHost...

### DIFF
--- a/notification/src/main/java/com/amazon/opendistroforelasticsearch/alerting/destination/client/DestinationHttpClient.java
+++ b/notification/src/main/java/com/amazon/opendistroforelasticsearch/alerting/destination/client/DestinationHttpClient.java
@@ -70,6 +70,7 @@ public class DestinationHttpClient {
                 .setDefaultRequestConfig(config)
                 .setConnectionManager(connectionManager)
                 .setRetryHandler(new DefaultHttpRequestRetryHandler())
+                .useSystemProperties()
                 .build();
     }
 


tomerilp:
Hi,
Is there any documentation on how to configure Slack notification with those proxy parameters?
Where exactly should I configure the proxy?

Thanks,

imranh2:
@tomerilp,

Always multiple ways to set proxy environment variables, but whatever suits you you 

I ended up setting it system wide for java /etc/java-8-openjdk/net.properties as we isolate the entire node to one role and can track every outgoing http request.

tomerilp:
> @tomerilp,
> 
> Always multiple ways to set proxy environment variables, but whatever suits you you
> 
> I ended up setting it system wide for java /etc/java-8-openjdk/net.properties as we isolate the entire node to one role and can track every outgoing http request.

I managed to make it work by adding `-Dhttps.proxyHost=proxy.domain.com -Dhttps.proxyPort=8080` to `/etc/sysconfig/elasticsearch`
I wonder if it possible to apply these settings only on the plugin scope so it won't affect the entire elasticsearch.

Thanks,

Issue #14, opened 2019-03-29
Throttle Period


Nishant23:
If we configure email or slack in the action block, it will send alerts each time monitor is triggered. In xpack of elasticseach this can be controlled by setting up `throttle_period` in action block. It will wait for the throttle_period amount of time after the first alert and then resend the alerts if the issue is not resolved yet once. Can we have the same functionality as in this?

For more info, you can look at https://www.elastic.co/guide/en/x-pack/current/actions.html#actions-ack-throttle

CarlMeadows:
Noted - thanks for the feedback. We do have the Acknowledge feature which will suspend subsequent notifications until the issue is resolved - but I can see folks also only wanting alerts to go out every 15 or 30 minutes on a 1 minute polling event - even before they can acknowledge it.

vamshin:
Adding my observations for the throttling feature. 

We could provide throttling at 3 levels. 
Level 1: Monitor
Level2: Trigger (If no trigger level throttling, it falls back to Monitor)
Level3: Action (If no Action level throttling, it falls back to trigger)

**Why would we need these levels?**
A Monitor could have multiple triggers. If we want all triggers to have similar throttling, then instead of adding throttling configuration at each trigger level, we could set at monitor level. Same logic holds for Trigger and Action 
Currently Xpack supports throttling at Trigger level and Action level. But For Xpack, 1 watch(Monitor) = 1 trigger, so trigger level throttling means similar to level 1 and level2 described above.

**Also should the throttling be global to all the alerts of the trigger or it should be with respect to a alert?**
If throttling is with respect to alert, then the throttling timer resets when alert is completed and the notification would be sent on next alert.
If throttling is global to alerts, then the timer runs across alerts.
One drawback we could think of with former approach is, if we have an alert thats going in and out, then the throttling does not come into picture.

 Please share feedback if you see any concerns or if i am missing something here

Nishant23:
When is this planned to be released?

elfisher:
@Nishant23 we are still working through how we want this mechanic to work. I'll be posting a high-level summary of our plans shortly to this issue thread. Stay tuned.

elfisher:
# [RFC] Alert throttling

The purpose of this request for comments (RFC) is to discuss how to enhance Open Distro for Elasticsearch Alerting to include throttling mechanisms on actions and provide users with the ability to undo acknowledgements. 

## Problem statement

[...]

PR #48, opened 2019-05-07
add throttle on action level

ylwu-amzn:
*Issue #14 *

*Description of changes:*
Add throttle support on Action level.

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/ActionExecutionResult.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/ActionExecutionResult.kt
@@ -52,7 +52,7 @@ data class ActionExecutionResult(
         @JvmStatic
         @Throws(IOException::class)
         fun parse(xcp: XContentParser): ActionExecutionResult {
-            var actionId: String? = null
+            lateinit var actionId: String
             var throttledCount: Int = 0
             var lastExecutionTime: Instant? = null
 
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/ActionExecutionResult.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/ActionExecutionResult.kt
@@ -1,0 +1,78 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.alerting.model
+
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.instant
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.optionalTimeField
+import org.elasticsearch.common.xcontent.ToXContent
+import org.elasticsearch.common.xcontent.ToXContentObject
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParserUtils
+import java.io.IOException
+import java.lang.IllegalStateException
+import java.time.Instant
+
+/**
+ * When an alert triggered, the trigger's actions will be executed.
+ * Action execution result records action throttle result and is a part of Alert.
+ */
+data class ActionExecutionResult(
+    val actionId: String,
+    val lastExecutionTime: Instant?,
+    val throttledCount: Int = 0
+) : ToXContentObject {
+
+    override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
+        return builder.startObject()
+                .field(ACTION_ID_FIELD, actionId)
+                .optionalTimeField(LAST_EXECUTION_TIME_FIELD, lastExecutionTime)
+                .field(THROTTLED_COUNT_FIELD, throttledCount)
+                .endObject()
+    }
+
+    companion object {
+        const val ACTION_ID_FIELD = "action_id"
+        const val LAST_EXECUTION_TIME_FIELD = "last_execution_time"
+        const val THROTTLED_COUNT_FIELD = "throttled_count"
+
+        @JvmStatic
+        @Throws(IOException::class)
+        fun parse(xcp: XContentParser): ActionExecutionResult {
+            var actionId: String? = null
+            var throttledCount: Int = 0
+            var lastExecutionTime: Instant? = null
+
+            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, xcp.currentToken(), xcp::getTokenLocation)
+            while (xcp.nextToken() != XContentParser.Token.END_OBJECT) {
+                val fieldName = xcp.currentName()
+                xcp.nextToken()
+                when (fieldName) {
+                    ACTION_ID_FIELD -> actionId = xcp.text()
+                    THROTTLED_COUNT_FIELD -> throttledCount = xcp.intValue()
+                    LAST_EXECUTION_TIME_FIELD -> lastExecutionTime = xcp.instant()
+
+                    else -> {
+                        throw IllegalStateException("Unexpected field: $fieldName, while parsing action")
+                    }
+                }
+            }
+
+            requireNotNull(actionId) { "Must set action id" }
+            return ActionExecutionResult(actionId, lastExecutionTime, throttledCount)
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -0,0 +1,76 @@
...
+        fun parse(xcp: XContentParser): Throttle {
+            var value: Int? = null
+            var unit: ChronoUnit? = null
+
+            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, xcp.currentToken(), xcp::getTokenLocation)
+            while (xcp.nextToken() != XContentParser.Token.END_OBJECT) {
+                val fieldName = xcp.currentName()
+                xcp.nextToken()
+                when (fieldName) {
+                    UNIT_FIELD -> unit = ChronoUnit.valueOf(xcp.text().toUpperCase(Locale.ROOT))
```
mihirsoni: we're supporting throttled in minutes unit currently, should we allow to have any different unit ? (i.e in Kibana we could control this by not passing any other units but in case of API one can add other valid units?)

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -0,0 +1,76 @@
...
+                    VALUE_FIELD -> {
+                        val currentToken = xcp.currentToken()
+                        when {
+                            currentToken == XContentParser.Token.VALUE_NULL -> value = null
+                            currentToken.isValue -> value = xcp.intValue()
```
mihirsoni: This Will also accept negative values, we shouldn't allow them.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
mihirsoni: [ChronoUnit.valueOf](https://docs.oracle.com/javase/10/docs/api/java/time/temporal/ChronoUnit.html#valueOf(java.lang.String)) throws NPE in case it is empty.  Could you verify that ?

bellfrog: We plan to support minutes unit only currently. Will add more checking.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
bellfrog: Yes, for negative values, it will not throttle. But it may confuse user. Will add negative checking.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
bellfrog: if null, throw error: null can not be a value of a non-null type String
if not a valid time unit, will throw java.lang.IllegalArgumentException: No enum constant

As we plan to support minutes unit only, will add logic to check the unit string value

core/src/main/resources/mappings/scheduled-jobs.json
```diff
@@ -114,6 +114,19 @@
                   "message_template": {
                     "type": "object",
                     "enabled": false
+                  },
+                  "throttle_enabled": {
```
dbbaughe: Any specific reason you have "throttle_enabled" outside "throttle"?
e.g.
```
"throttle": {
    "properties": {
        "enabled": { "type": "boolean" },
        "value": { "type": "integer" },
        "unit": { "type": "keyword": }
    }
}
```

ylwu-amzn: We may add throttle on Monitor/Trigger level, they function as default throttle if the lower level's throttle is not defined. I think the throttle enable or not should be on action level. That's why put the "throttle_enabled" outside "throttle".

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.elasticsearch.common.xcontent.XContentBuilder
+import org.elasticsearch.common.xcontent.XContentParser
+import org.elasticsearch.common.xcontent.XContentParserUtils
+import java.io.IOException
+import java.lang.IllegalStateException
+import java.time.temporal.ChronoUnit
+import java.util.Locale
+
+data class Throttle(
+    val value: Int?,
```
dbbaughe: Why is value nullable?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -0,0 +1,84 @@
...
+                        }
+                    }
+
+                    else -> {
+                        throw IllegalStateException("Unexpected field: $fieldName, while parsing action")
+                    }
+                }
+            }
+
+            return if (unit == null) Throttle(value = value) else Throttle(value = value, unit = unit)
```
dbbaughe: This allows someone to do { throttle: { value: 1 } }, why not require unit? I see it'll default to MINUTEs when not specified, but why even allow it?

ylwu-amzn: Here the unit is kept for the flexibility to support more options like HOURS

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -0,0 +1,84 @@
...
+            var unit: ChronoUnit? = null
+
+            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, xcp.currentToken(), xcp::getTokenLocation)
+            while (xcp.nextToken() != XContentParser.Token.END_OBJECT) {
+                val fieldName = xcp.currentName()
+                xcp.nextToken()
+                when (fieldName) {
+                    UNIT_FIELD -> {
+                        val unitString = xcp.text().toUpperCase(Locale.ROOT)
+                        require(StringUtils.equals(unitString, ChronoUnit.MINUTES.name), { "Only support MINUTES throttle unit" })
```
dbbaughe: Why are we only supporting MINUTES for initial release of this feature?
Seems like unnecessary burden on users to do conversion to minutes when dealing with monitors with longer frequencies.

ylwu-amzn: As the throttle is just to avoid too many notifications in a short time, so we support MINUTES first. It does bring extra to do conversion to minutes. We have plan to support more units based on user's feedback.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
ylwu-amzn: Will change it to non-nullable.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/ActionExecutionResult.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    companion object {
+        const val ACTION_ID_FIELD = "action_id"
+        const val LAST_EXECUTION_TIME_FIELD = "last_execution_time"
+        const val THROTTLED_COUNT_FIELD = "throttled_count"
+
+        @JvmStatic
+        @Throws(IOException::class)
+        fun parse(xcp: XContentParser): ActionExecutionResult {
+            var actionId: String? = null
```
dbbaughe: I believe we have been using `lateinit var actionId: String` instead of `String?` elsewhere

ylwu-amzn: Thanks, will change it

core/src/main/resources/mappings/scheduled-jobs.json
dbbaughe: But, will we also support "enabled" at the trigger and monitor level? Or will "enabled" only be at the action level and eventually action, trigger, monitors will have "throttle"?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
dbbaughe: I'm confused why requiring unit explicitly would stop us from allowing more options like HOURS in the future?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
dbbaughe: Is it a big change to support more units other than just MINUTES? If not I don't see why we are not supporting something like MINUTES, HOURS, DAYS from the start rather than waiting for a user to open an issue asking for us to support it. It looks like a single line change right here?

core/src/main/resources/mappings/scheduled-jobs.json
dbbaughe: From offline discussion: It's to make falling back to higher levels easy

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
dbbaughe: From offline discussion: While not a big change for backend, it requires more significant UI change. Will release with this to start and iteratively increase to more units later.

PR #59, opened 2019-05-17
Throttle

ylwu-amzn:
1.Fix update action execution result bug;
2.Add integration test code change

### DIFF
--- a/alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
+++ b/alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.alerting
 
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertError
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertIndices
+import com.amazon.opendistroforelasticsearch.alerting.core.model.IntervalSchedule
 import com.amazon.opendistroforelasticsearch.alerting.model.Alert
 import com.amazon.opendistroforelasticsearch.alerting.model.Alert.State.ACKNOWLEDGED
 import com.amazon.opendistroforelasticsearch.alerting.model.Alert.State.ACTIVE
@@ -24,6 +25,8 @@ import com.amazon.opendistroforelasticsearch.alerting.model.Alert.State.COMPLETE
 import com.amazon.opendistroforelasticsearch.alerting.model.Alert.State.ERROR
 import com.amazon.opendistroforelasticsearch.alerting.model.Monitor
 import com.amazon.opendistroforelasticsearch.alerting.core.model.SearchInput
+import com.amazon.opendistroforelasticsearch.alerting.model.ActionExecutionResult
+import com.amazon.opendistroforelasticsearch.alerting.model.action.Throttle
 import org.elasticsearch.common.settings.Settings
 import org.elasticsearch.index.query.QueryBuilders
 import org.elasticsearch.rest.RestStatus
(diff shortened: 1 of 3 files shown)


alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -507,6 +512,97 @@ class MonitorRunnerIT : AlertingRestTestCase() {
...
+        for (result in alert.actionExecutionResults) {
+            val expected = expectedResult[result.actionId]
+            assertEquals(expected, result.throttledCount)
+            actionResult.put(result.actionId, result)
+        }
+        return actionResult
+    }
+
+    private fun verifyActionThrottleResults(monitoId: String, expectedResult: Map<String, Boolean>) {
+        val output = entityAsMap(executeMonitor(monitoId))
```
mihirsoni: Could you move this where this function is being called. Because name says verify but it also executeMonitor.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -507,6 +512,97 @@ class MonitorRunnerIT : AlertingRestTestCase() {
...
+                Pair(actionThrottleNotEnabled.id, false)))
+
+        val alerts1 = searchAlerts(monitor)
+        assertEquals("1 alert should be returned", 1, alerts1.size)
+        verifyAlert(alerts1.single(), monitor, ACTIVE)
+        val actionResults1 = verifyActionExecutionResultInAlert(alerts1[0],
+                mutableMapOf(Pair(actionThrottleEnabled.id, 0), Pair(actionThrottleNotEnabled.id, 0)))
+
+        assertEquals(actionResults1.size, 2)
+        verifyActionThrottleResults(monitor.id, mutableMapOf(Pair(actionThrottleEnabled.id, true),
```
mihirsoni: Let's move executeMonitor on here so that it is easy to understand.

ylwu-amzn: Thanks, change it

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -507,6 +512,97 @@ class MonitorRunnerIT : AlertingRestTestCase() {
...
+                destinationId = createDestination().id,
+                throttleEnabled = false, throttle = Throttle(value = 5, unit = MINUTES))
+        val actions = listOf(actionThrottleEnabled, actionThrottleNotEnabled)
+        val monitor = createMonitor(randomMonitor(triggers = listOf(randomTrigger(condition = ALWAYS_RUN, actions = actions)),
+                schedule = IntervalSchedule(interval = 1, unit = ChronoUnit.MINUTES)))
+
+        verifyActionThrottleResults(monitor.id, mutableMapOf(Pair(actionThrottleEnabled.id, false),
+                Pair(actionThrottleNotEnabled.id, false)))
+
+        val alerts1 = searchAlerts(monitor)
```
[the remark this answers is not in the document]
ylwu-amzn: Thanks ,change it

dbbaughe:
Unrelated to this specific PR, but:
Adding Throttling introduced an update to our mappings. We need to add "schema_version" to the mappings and set it to 1 with this update. And in the code if schema_version does not exist on the document we'll assume it's 0 (ie initial release/before throttling was added).

We also need to add a way to update the mappings on each cluster. Currently none of the index mappings will be updated since we only do that once on initial use of alerting.

All of this needs to happen ASAP before we cut a release for 7.0

ylwu-amzn:
Create an new issue to track the schema versioning. https://github.com/opendistro-for-elasticsearch/alerting/issues/60

Issue #60, opened 2019-05-20
Index mapping schema versioning


ylwu-amzn:
Unrelated to this specific PR, but:
Adding Throttling introduced an update to our mappings. We need to add "schema_version" to the mappings and set it to 1 with this update. And in the code if schema_version does not exist on the document we'll assume it's 0 (ie initial release/before throttling was added).

We also need to add a way to update the mappings on each cluster. Currently none of the index mappings will be updated since we only do that once on initial use of alerting.

All of this needs to happen ASAP before we cut a release for 7.0

_Originally posted by @dbbaughe in https://github.com/opendistro-for-elasticsearch/alerting/pull/59#issuecomment-493647105_

ylwu-amzn:
Code change merged.

PR #61, opened 2019-05-23
index mapping schema versioning

ylwu-amzn:
*Issue #60*

*Description of changes:*
Add index mapping version, update current index mapping if new version published.


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -184,6 +192,16 @@ class AlertIndices(
         }
     }
 
+    private fun updateIndexMapping(index: String, indexPrefix: Boolean = false) {
+        val clusterState = clusterService.state()
+        when {
+            indexPrefix -> clusterState.metaData.indices.keysIt().forEach {
+                key -> if (key.startsWith(index)) SchemaVersionUtils.updateIndexMapping(key, MAPPING_TYPE, alertMapping(),
```
lucaswin-amzn: I assume this is for updating all the history indexes? Is this needed or will updating the mappings for `*alerts*` update all of them in one go?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexDestinationAction.kt
```diff
@@ -117,6 +117,7 @@ class RestIndexDestinationAction(
             if (!scheduledJobIndices.scheduledJobIndexExists()) {
                 scheduledJobIndices.initScheduledJobIndex(ActionListener.wrap(::onCreateMappingsResponse, ::onFailure))
             } else {
+                scheduledJobIndices.updateScheduledJobIndex()
```
lucaswin-amzn: While I am sure this solution works. Rather than performing this request on every indexing operation of monitors / destinations it would be great to check if the schema version has changed, if it hasn't there is no need to perform the request.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -507,6 +512,96 @@ class MonitorRunnerIT : AlertingRestTestCase() {
         assertEquals("Enabled times do not match", monitor.enabledTime, retrievedMonitor.enabledTime)
     }
 
+    fun `test monitor with throttled action for same alert`() {
```
lucaswin-amzn: Throttling ITs should go into another PR. This PR is specific to schema mappings.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
```diff
@@ -0,0 +1,86 @@
...
+                        }
+                    }
+                }
+                xcp.nextToken()
+            }
+            return 0
+        }
+
+        @JvmStatic
+        @Suppress("UNCHECKED_CAST")
```
lucaswin-amzn: Anyway around suppressing these unchecked casts? This could lead to future bugs.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
```diff
@@ -0,0 +1,86 @@
...
+        fun shouldUpdateIndex(index: IndexMetaData, mapping: String): Boolean {
+            var oldVersion = 0
```
mihirsoni: Just a thought : Do we want old version always to be 0 ? For example we could have situation in future that newer version i.e 2.0 is not directly upgradable from 0. So you would have to do migration path could be 0->1->2 ?  This is not an issue right now but we should keep in mind this or add a comment on our current decision.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
ylwu-amzn: Update all alerts history index mapping to keep consistent. It's easier for user to use same mapping rather than different versions.
Will check the schema version and update the mapping if it's not the latest one. This will only happen once for each index.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexDestinationAction.kt
ylwu-amzn: Yes, agree. Actually, I have put the checking logic in "SchemaVersionUtils.shouldUpdateIndex", if the schema version not changed, will skip updating the index mapping.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
ylwu-amzn: My bad, please ignore these changes and review other parts first.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
ylwu-amzn: Agree, will check it

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
ylwu-amzn: As we have no index mapping schema versioning before, regard the default version as 0. If the index has schema version info, will update oldVersion to its version: "if (metaData.containsKey(SCHEMA_VERSION)) oldVersion = metaData[SCHEMA_VERSION] as Int"
The migration path could be 0->1->1->2, or just 0->2.

ylwu-amzn: Offline discussed, we can define something to control effective migration path, such as 0->1 is valid , but 0->2 is not. Currently, we use simple migration logic: replace all old version with latest version. Will consider migration control in future.

dbbaughe:
From offline discussion:
1. We need schema_version on the documents themselves, not just the index.
2. Schema_version on the documents should be controlled/updated by us, not the user.
3. We do not need to update rolled over alert history indices with new mappings since they won't be updated with your new properties.
4. We should not update these in monitor runner/index API. Move the update check to plugin/node startup.

ylwu-amzn:
From more offline discuss, will do such changes
1. Shorten checking logic by adding variable to record index updated or not
2.Add schema_version on documents and control the document's schema version update
3.Don't update rolled over history indices. Just update

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
```diff
@@ -0,0 +1,85 @@
...
+                        }
+                    }
+                }
+                xcp.nextToken()
+            }
+            return 0
+        }
+
+        @JvmStatic
+        fun shouldUpdateIndex(index: IndexMetaData, mapping: String): Boolean {
```
vamshi1252: Better to make the **oldversion** as parameter in function.  So tomorrow to when old version becomes 1, we do not have to change this function

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -248,22 +250,29 @@ class MonitorRunner(
...
+            alertSchemaVersion = SchemaVersionUtils.getSchemaVersion(AlertIndices.alertMapping())
+        }
+
         // Merge the alert's error message to the current alert's history
         val updatedHistory = currentAlert?.errorHistory.update(alertError)
         return if (alertError == null && !result.triggered) {
             currentAlert?.copy(state = COMPLETED, endTime = currentTime, errorMessage = null,
-                    errorHistory = updatedHistory, actionExecutionResults = updatedActionExecutionResults)
+                    errorHistory = updatedHistory, actionExecutionResults = updatedActionExecutionResults,
+                    schemaVersion = alertSchemaVersion!!)
```
lucaswin-amzn: Can we avoid using `!!` here? and on line 269/275?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -151,16 +156,20 @@ class AlertIndices(
...
         }
         alertIndexInitialized
     }
 
-    suspend fun createInitialHistoryIndex() {
+    suspend fun createOrUpdateInitialHistoryIndex() {
         if (!historyIndexInitialized) {
             historyIndexInitialized = createIndex(HISTORY_INDEX_PATTERN, HISTORY_WRITE_INDEX)
+        } else {
+            updateIndexMapping(HISTORY_WRITE_INDEX, true)
```
lucaswin-amzn: Shouldn't this have the same check as `createorUpdateAlertIndex`?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -184,6 +193,34 @@ class AlertIndices(
...
+            return
+        }
+
+        var putMappingRequest: PutMappingRequest = PutMappingRequest(targetIndex).type(MAPPING_TYPE)
+                .source(mapping, XContentType.JSON)
+        val updateResponse: AcknowledgedResponse = client.suspendUntil { client.putMapping(putMappingRequest, it) }
+        if (updateResponse.isAcknowledged) {
+            logger.info("Index mapping of $targetIndex is updated")
+            setIndexUpdateFlag(index, targetIndex)
+        }
```
lucaswin-amzn: Something should be done in an error case. At the very least add a log statement.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexDestinationAction.kt
```diff
@@ -145,6 +155,18 @@ class RestIndexDestinationAction(
...
 
+        private fun onUpdateMappingsResponse(response: AcknowledgedResponse) {
+            if (response.isAcknowledged) {
+                log.info("Updated  ${ScheduledJob.SCHEDULED_JOBS_INDEX} with mappings.")
+                scheduledJobIndexUpdated = true
+                prepareDestinationIndexing()
+            } else {
+                log.error("Update ${ScheduledJob.SCHEDULED_JOBS_INDEX} mappings call not acknowledged.")
+                channel.sendResponse(BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR,
+                        response.toXContent(channel.newErrorBuilder(), ToXContent.EMPTY_PARAMS)))
```
lucaswin-amzn: This will give a blank response back to the user. We should construct a basic error message with the `channel.newErrorBuilder()`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -174,7 +193,21 @@ class RestIndexMonitorAction(
...
 
+        private fun onUpdateMappingsResponse(response: AcknowledgedResponse) {
+            if (response.isAcknowledged) {
+                log.info("Updated ${ScheduledJob.SCHEDULED_JOBS_INDEX} with mappings.")
+                scheduledJobIndexUpdated = true
+                prepareMonitorIndexing()
+            } else {
+                log.error("Updated ${ScheduledJob.SCHEDULED_JOBS_INDEX} mappings call not acknowledged.")
+                channel.sendResponse(BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR,
+                        response.toXContent(channel.newErrorBuilder(), EMPTY_PARAMS)))
```
lucaswin-amzn: Same thing here. Lets provide an error message.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertingRestTestCase.kt
```diff
@@ -111,6 +112,19 @@ abstract class AlertingRestTestCase : ESRestTestCase() {
                 customWebhook = null)
     }
 
+    @Suppress("UNCHECKED_CAST")
```
lucaswin-amzn: I think you can avoid suppressing this by using the `.stringMap()`(i believe that's what it is called) function from the test helpers.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndicesIT.kt
```diff
@@ -33,7 +34,28 @@ class AlertIndicesIT : AlertingRestTestCase() {
...
+
+        putAlertMappings(AlertIndices.alertMapping().trimStart('{').trimEnd('}')
+                .replace("\"schema_version\": 1", "\"schema_version\": 0"))
+        assertIndexExists(AlertIndices.ALERT_INDEX)
+        assertIndexExists(AlertIndices.HISTORY_WRITE_INDEX)
+        verifyIndexSchemaVersion(AlertIndices.ALERT_INDEX, 0)
+        verifyIndexSchemaVersion(AlertIndices.HISTORY_WRITE_INDEX, 0)
+
+        executeMonitor(createRandomMonitor())
+        Thread.sleep(2000)
```
lucaswin-amzn: Two things here:
1. We should avoid sleeps in tests as these will add up over time and cause testing to become very long.
2. Since we already executed the monitor an alert should be created?

Or am I understanding the purpose of this sleep wrong?

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
```diff
@@ -0,0 +1,99 @@
...
+    companion object {
+        const val _META = "_meta"
+        const val SCHEMA_VERSION = "schema_version"
+
+        @JvmStatic
+        fun getSchemaVersion(mapping: String): Int {
+            val xcp = XContentType.JSON.xContent().createParser(NamedXContentRegistry.EMPTY,
+                    LoggingDeprecationHandler.INSTANCE, mapping)
+
+            while (!xcp.isClosed) {
```
lucaswin-amzn: This parser appears to be different than the other model classes. You might want to use the same style here, as this logic looks rather complex and could be simplified.

core/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtilsTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        val schemaVersion = SchemaVersionUtils.getSchemaVersion(message)
+        assertEquals(0, schemaVersion)
+    }
+
+    @Test
+    fun `test get schema version without schema_version`() {
+        val message = "{\"user\":{ \"name\":\"test\"},\"_meta\":{\"test\": 1}}"
+
+        val schemaVersion = SchemaVersionUtils.getSchemaVersion(message)
+        assertEquals(0, schemaVersion)
```
lucaswin-amzn: We should have a test for invalid / negative schema versions.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
ylwu-amzn: For alert history index, the index will change according to current time. So even alert config index is updated, the alert history write index may still use old schema.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
ylwu-amzn: Sure, will add log for error case

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexDestinationAction.kt
ylwu-amzn: Will change it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ylwu-amzn: Will change it

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -248,22 +250,29 @@ class MonitorRunner(
                     if (it.value.throttled) 1 else 0) })
         }
 
+        if (alertSchemaVersion == null) {
```
jinsoor-amzn: I don't see any other places that uses `alertSchemaVersion` out side of this function and this if-statement will always be true because of that. Can we keep `alertSchemaVersion` scope within the function and this can also help to address lucaswin's comment about using `!!`

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/Monitor.kt
```diff
@@ -47,6 +47,7 @@ data class Monitor(
     override val schedule: Schedule,
     override val lastUpdateTime: Instant,
     override val enabledTime: Instant?,
+    val schemaVersion: Int = 0,
```
jinsoor-amzn: There is many other places that require 0 as default schema version. Lets create a constant with  `NO_SCHEMA_VERSION` to use as default.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -78,10 +80,23 @@ class RestIndexMonitorAction(
 ) : BaseRestHandler(settings) {
 
     private var scheduledJobIndices: ScheduledJobIndices
+
     @Volatile private var maxMonitors = ALERTING_MAX_MONITORS.get(settings)
     @Volatile private var requestTimeout = REQUEST_TIMEOUT.get(settings)
     @Volatile private var indexTimeout = INDEX_TIMEOUT.get(settings)
 
+    companion object {
+        var scheduledJobSchemaVersion: Int? = null
```
jinsoor-amzn: with this other classes can set values to this variable. You want to add `private set` to prevent it from publicly modifiable. Kotlin syntax should be 

```kotlin
var scheduledJobSchemaVersion: Int? = null
    private set
```

https://kotlinlang.org/docs/reference/properties.html

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -78,10 +80,23 @@ class RestIndexMonitorAction(
...
+        var scheduledJobIndexUpdated: Boolean = false
```
jinsoor-amzn: Same here. You should add `private set` to prevent from publicly modifiable.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
```diff
@@ -0,0 +1,99 @@
...
+                            if (xcp.currentName() == SCHEMA_VERSION) {
+                                return xcp.intValue()
+                            }
+                            xcp.nextToken()
+                        }
+                    }
+                }
+                xcp.nextToken()
+            }
+            return 0
```
jinsoor-amzn: I assume `return 0` is supposed to be the default schema version. Please update this to also make use of the `NO_SCHEMA_VERSION` constant.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ylwu-amzn: If move `alertSchemaVersion` within the function, have to get `alertSchemaVersion` for every `composeAlert` call. So prefer to keep it as class property.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/Monitor.kt
ylwu-amzn: Sure, will change it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ylwu-amzn: Good suggestion. Will add `private set`

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
ylwu-amzn: Will change it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ylwu-amzn: `scheduledJobIndexUpdated` is used by `RestIndexDestinationAction.onUpdateMappingsResponse`. When a destination or monitor is updated, will update alerting config index and set `scheduledJobIndexUpdated = true`

If add `private set`, `RestIndexDestinationAction` can't set `scheduledJobIndexUpdated` any more

core/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtilsTest.kt
ylwu-amzn: Will add it

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
ylwu-amzn: You mean use `while (xcp.nextToken() != Token.END_OBJECT) {` rather than  `while (!xcp.isClosed) {` ? I will not work as `xcp` has multiple `END_OBJECT` tokens and only `while (!xcp.isClosed) {` can iterate every object rather than stop at `END_OBJECT` token.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertingRestTestCase.kt
ylwu-amzn: Change it.

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndicesIT.kt
ylwu-amzn: Will remove it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
jinsoor-amzn: After reading the PR again, we can modify SchemaVersionUtils a bit to make it more immutable.

`SchemaVersionUtils` should have functions to get schemaversion to get specific index version. ie. `SchemaVersionUtils.getAlertIndexSchemaVersion()` and `SchemaVersionUtils.getScheduledJobSchemaVersion()`. In those functions, we can hide this logic of initializing once if such schema version does not exist. This way if in future other class needs `AlertIndexSchemaVersion` or `ScheduledJobSchemaVersion` that class does not need to build this complicated logic.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ylwu-amzn: Will change it

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
jinsoor-amzn: This sounds like a logic that should live in `SchemaVersionUtils`. Something like `SchemaVersionUtils.isScheduledJobIndexUpdated()` which internally does the updating logic so it can be shared between `IndexMonitorAction` and IndexDestinationAction`.

In addition, in case the index was created with newer Schema how will this be set to `true`? In this case will we do needless update on the ScheduledJob index because `scheduledJobIndexUpdated` was initialized to `false`?

jinsoor-amzn:
In addition to making `SchemaVersionUtils` more immutable. Do you think it is work it to make the `SchemaVersionUtils` into a singleton object `SchemaVersionService`? With the additional logic that the `SchemaVersionUtils`  will have it may require a ESClient object and I think it make sense to have some constructor that does initial loading of SchemaVersions .

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/util/SchemaVersionUtils.kt
lucaswin-amzn: Then lets remove the if statements and move to a when statement:

`
val fieldName = xcp.currentName()
xcp.nextToken()
when (fieldName):
     SCHEMA_VERSION -> {}
     else -> xcp.skipChildren()
`

Or something along these lines.

ylwu-amzn: Ok, will change it

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ylwu-amzn: SchemaVersionUtils is in `core` package now, so it can't access alerting index as it's in `alerting` package. 
Will add extra util class in alerting package and initialize alert and scheduled job index schema version set

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ylwu-amzn: Same as last comment, will add a new util class

ylwu-amzn:
Post the build result of `./gradlew build` on my laptop: 

> Configure project :alerting
  Gradle Version        : 5.2.1
  OS Info               : Mac OS X 10.13.6 (x86_64)
  JDK Version           : 12 (Oracle Corporation 12 [Java HotSpot(TM) 64-Bit Server VM 12+33])
  JAVA_HOME             : /Library/Java/JavaVirtualMachines/jdk-12.jdk/Contents/Home
  Random Testing Seed   : C76C9762AE562896

BUILD SUCCESSFUL in 2m 44s

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/IndexUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.alerting.util
+
+import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertIndices
+import com.amazon.opendistroforelasticsearch.alerting.core.ScheduledJobIndices
+import com.amazon.opendistroforelasticsearch.alerting.core.util.SchemaVersionUtils.Companion.getSchemaVersion
+
+class IndexUtils {
```
jinsoor-amzn: Please combine `SchemaVersionUtils` in `core` package into here.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ylwu-amzn: Missed some comments "in case the index was created with newer Schema how will this be set to true?", will set index updated as true if index is newly created

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/IndexUtils.kt
ylwu-amzn: ok, will merge it

PR #67, opened 2019-06-11
Maximum limit for Throttle.

ylwu-amzn:
Add limit for action throttle, max throttle period is one day

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
@@ -41,6 +41,7 @@ data class Throttle(
     companion object {
         const val VALUE_FIELD = "value"
         const val UNIT_FIELD = "unit"
+        const val MAX_THROTTLE_VALUE = 1440 // 1440 minutes = one day
 
         @JvmStatic
         @Throws(IOException::class)
@@ -65,6 +66,7 @@ data class Throttle(
                             currentToken.isValue -> {
                                 value = xcp.intValue()
                                 require(value > 0, { "Can only set positive throttle period" })
+                                require(value < MAX_THROTTLE_VALUE, { "Can only set throttle period less than 1 day" })
                             }
                             else -> {
                                 XContentParserUtils.throwUnknownToken(currentToken, xcp.tokenLocation)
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -65,6 +66,7 @@ data class Throttle(
                             currentToken.isValue -> {
                                 value = xcp.intValue()
                                 require(value > 0, { "Can only set positive throttle period" })
+                                require(value < MAX_THROTTLE_VALUE, { "Can only set throttle period less than 1 day" })
```
mihirsoni: I see on kibana max is 1440, so if user pass on 1440 it will fail here. Can we make it <= ?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
```diff
@@ -41,6 +41,7 @@ data class Throttle(
     companion object {
         const val VALUE_FIELD = "value"
         const val UNIT_FIELD = "unit"
+        const val MAX_THROTTLE_VALUE = 1440 // 1440 minutes = one day
```
jinsoor-amzn: Lets make sure this can also work when we introduce new Unit. Can you add validation in the constructor of `Throttle` or before creating the `Throttle` Object so you have both Value and Unit when checking for max throttle value?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
ylwu-amzn: Good point, will change it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt
ylwu-amzn: We only support MINUTES unit currently, already has checking now.

require(StringUtils.equals(unitString, ChronoUnit.MINUTES.name), { "Only support MINUTES throttle unit currently" })

PR #69, opened 2019-06-13
configure action throttle max value in ES setting

ylwu-amzn:
*configure action throttle max value in ES setting*


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -153,6 +158,7 @@ class RestIndexMonitorAction(
          * and compare this to the [maxMonitorCount]. Requests that breach this threshold will be rejected.
          */
         private fun prepareMonitorIndexing() {
+            Validator.validateActionThrottle(newMonitor, maxActionThrottle, TimeValue.timeValueMinutes(1))
```
jinsoor-amzn: What is the reason for creating a separate `Validator` class when this validate is only being used in this RestHandler? I think just having the `validate` as part of a function in this RestHandler is simpler.

ylwu-amzn: 1.We can move other validation logic to this class to manage the validate logic easily.
2.Separate the validation logic and we can write UT easily.

jinsoor-amzn: 1. This `Validator` class gives will give impression that this validator. However for `validateActionThrottle` this validation is only being used in this RestHandler and are not being used else where. Having the `validateActionThrottle` in this RestHandler as a private method will give more contained usage until we find this logic is going to be shared somewhere else.

2. We already have integ test in `MonitorRestApiIT` that tests against ES cluster. We should add to test case there and ensure right errors are thrown back from the RestClient.

ylwu-amzn: 1.It's not a solid reason to put a method in one class as private just as it's only used in this class currently. That will cause class inflation and not a readable style to put many logics in one class. 
2.Will add more IT cases. But I think we should add more UT cases too and split the validation logic can help us to write UT more easily.

jinsoor-amzn: 1. It is also not a solid reason create a new class with single function when it is currently only being used in single place. In addition, `Validator` is not very intuitive name to give a class. We have many other validating that is happening through Alerting and it is not like we are going to put all the validation function into this class. If the concern is the size of this RestHandler getting to large, there is better candidate to put this validation function instead of creating this new class. I would say [Throttle.kt](https://github.com/opendistro-for-elasticsearch/alerting/blob/master/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/action/Throttle.kt) is a good place to this validation. `Throttle.kt` already have `companion object` to easily add the new function, it fits well with the main goal for the class to contain everything related to Throttle, and it does not create unnecessary single function class definition.

2. In addition to having UT, IT in this case is necessarily. Exception/Error messages that need to be relayed back to client side are not effective when tested in UT. In general, I have seen many times where the function throws a correct exception but the RestClient/RestHandler handles the exception poorly and ultimately throwing incorrect error message back to client. For this reason, it is best for us to test API error message using the RestHandler in our ESIntegTest to closely replicate the end to end use case.

ylwu-amzn: hi, Jinsoo, thanks for your comments. To make the Throttle only owns model role,  how about we still put the validate method to RestIndexMonitorAction as private method and I will remove the UT cases as you said the UT may not work sometimes. Will add IT cases to cover such UT?

ylwu-amzn:
Starting a Gradle Daemon (subsequent builds will be faster)

> Configure project :alerting
  Gradle Version        : 5.2.1
  OS Info               : Mac OS X 10.13.6 (x86_64)
  JDK Version           : 12 (Oracle Corporation 12 [Java HotSpot(TM) 64-Bit Server VM 12+33])
  JAVA_HOME             : /Library/Java/JavaVirtualMachines/jdk-12.jdk/Contents/Home
  Random Testing Seed   : 918EB3F52067FEA0

BUILD SUCCESSFUL in 3m 19s

Issue #47, opened 2019-05-03
Support more input types other than the Elasticsearch DSL queries


dbbaughe:
Our initial release supports just the Elasticsearch DSL queries.

We have started getting more requests for other input types or enhancements.
Cluster Health API: https://discuss.opendistrocommunity.dev/t/cluster-health-alert-monitors/606
Input Chaining: https://github.com/opendistro-for-elasticsearch/alerting/issues/43

ShahNewazKhan:
Other use-cases also include `indices api` such as `stats / segments / recovery / shard stores` etc.

qreshi:
Closing in favor of https://github.com/opensearch-project/alerting/issues/39

PR #82, opened 2019-07-16
Adding new type of input for Monitors - HttpInput

ann3431:
*Issue #, if available: #47 

*Description of changes: 

- Adding new type of input ```HttpInput``` for monitors. 

```MonitorRunner``` class creates a ```HttpInputClient``` when collecting input results of an ```HttpInput```. 

- Changes to ```README``` file regarding setting ```JAVA_HOME```.

```
> > Configure project :alerting
  Gradle Version        : 5.2.1
  OS Info               : Mac OS X 10.14.5 (x86_64)
  JDK Version           : 12 (Oracle Corporation 12.0.1 [Java HotSpot(TM) 64-Bit Server VM 12.0.1+12])
  JAVA_HOME             : /Library/Java/JavaVirtualMachines/jdk-12.0.1.jdk/Contents/Home
  Random Testing Seed   : 9087A73AD72ABD1

BUILD SUCCESSFUL in 1m 49s
52 actionable tasks: 30 executed, 22 up-to-date
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ Please see our [documentation](https://opendistro.github.io/for-elasticsearch-do
 1. Check out this package from version control.
 1. Launch Intellij IDEA, choose **Import Project**, and select the `settings.gradle` file in the root of this package. 
 1. To build from the command line, set `JAVA_HOME` to point to a JDK >= 12 before running `./gradlew`.
+  - Unix System
+    1. `export JAVA_HOME=jdk-install-dir`: Replace `jdk-install-dir` by the JAVA_HOME directory of your system.
+    1. `export PATH=$JAVA_HOME/bin:$PATH`
+ 
+  - Windows System
+    1. Find **My Computers** from file directory, right click and select **properties**.
+    1. Select the **Advanced** tab, select **Environment variables**.
+    1. Edit **JAVA_HOME** to path of where JDK softeare is installed.
 
 
 ## Build
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -49,7 +49,10 @@ configurations.all {
         force "commons-logging:commons-logging:${versions.commonslogging}"
         force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
         force "commons-codec:commons-codec:${versions.commonscodec}"
-        
+        force "commons-collections:commons-collections:3.2.2"
+        force "org.apache.httpcomponents:httpcore-nio:4.4.11"
+        force "org.apache.httpcomponents:httpclient:4.5.7"
+
         // This is required because kotlin coroutines core-1.1.1 still requires kotin stdlib 1.3.20 and we're using 1.3.21
         force "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
         force "org.jetbrains.kotlin:kotlin-stdlib-common:${kotlin_version}"
@@ -58,6 +61,7 @@ configurations.all {
 
 dependencies {
     compileOnly "org.elasticsearch.plugin:elasticsearch-scripting-painless-spi:${versions.elasticsearch}"
+    compile "org.apache.httpcomponents:httpasyncclient:4.1.4"
 
     // Elasticsearch Nanny state
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -111,7 +111,7 @@ class MonitorRunner(
 ) : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
     private val logger = LogManager.getLogger(MonitorRunner::class.java)
-    private var httpClient: HttpInputClient
+    private lateinit var httpClient: HttpInputClient
     private lateinit var runnerSupervisor: Job
     override val coroutineContext: CoroutineContext
         get() = Dispatchers.Default + runnerSupervisor
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -111,7 +111,7 @@ class MonitorRunner(
 ) : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
     private val logger = LogManager.getLogger(MonitorRunner::class.java)
-    private lateinit var httpClient: HttpInputClient
+    private var httpClient: HttpInputClient
     private lateinit var runnerSupervisor: Job
     override val coroutineContext: CoroutineContext
         get() = Dispatchers.Default + runnerSupervisor
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -18,7 +18,12 @@ package com.amazon.opendistroforelasticsearch.alerting
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertError
 import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertIndices
 import com.amazon.opendistroforelasticsearch.alerting.alerts.moveAlerts
+import com.amazon.opendistroforelasticsearch.alerting.client.HttpInputClient
 import com.amazon.opendistroforelasticsearch.alerting.core.JobRunner
+import com.amazon.opendistroforelasticsearch.alerting.core.httpapi.suspendUntil
+import com.amazon.opendistroforelasticsearch.alerting.core.httpapi.toGetRequest
+import com.amazon.opendistroforelasticsearch.alerting.core.httpapi.toMap
+import com.amazon.opendistroforelasticsearch.alerting.core.model.HttpInput
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob.Companion.SCHEDULED_JOBS_INDEX
 import com.amazon.opendistroforelasticsearch.alerting.core.model.SearchInput
@@ -51,13 +56,14 @@ import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.MOVE_ALERTS_BACKOFF_COUNT
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.MOVE_ALERTS_BACKOFF_MILLIS
 import com.amazon.opendistroforelasticsearch.alerting.util.IndexUtils
-import org.apache.logging.log4j.LogManager
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import org.apache.http.HttpResponse
+import org.apache.logging.log4j.LogManager
 import org.elasticsearch.ExceptionsHelper
 import org.elasticsearch.action.DocWriteRequest
 import org.elasticsearch.action.bulk.BackoffPolicy
@@ -292,7 +298,14 @@ class MonitorRunner(
                         val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
                         results += searchResponse.convertToMap()
                     }
-                    else -> {
+                    is HttpInput -> {
+                        val httpClient = HttpInputClient()
+                        httpClient.client.start()
+                        val response: HttpResponse = httpClient.client.suspendUntil {
+                            httpClient.client.execute(input.toGetRequest(), it) }
+                        httpClient.client.close()
+                        results += response.toMap()
+                    } else -> {
                         throw IllegalArgumentException("Unsupported input type: ${input.name()}.")
                     }
                 }
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
@@ -1,0 +1,55 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.alerting.client
+
+import com.amazon.opendistroforelasticsearch.alerting.core.model.HttpInput
+import org.apache.http.client.config.RequestConfig
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder
+import org.elasticsearch.common.unit.TimeValue
+import java.security.AccessController
+import java.security.PrivilegedAction
+
+/**
+ * This class takes [HttpInput] and performs GET requests to given URIs.
+ */
+class HttpInputClient {
+
+    // TODO: If possible, these settings should be implemented as changeable via the "_cluster/settings" API.
+    private val CONNECTION_TIMEOUT_MILLISECONDS = TimeValue.timeValueSeconds(10).millis().toInt()
+    private val REQUEST_TIMEOUT_MILLISECONDS = TimeValue.timeValueSeconds(10).millis().toInt()
+    private val SOCKET_TIMEOUT_MILLISECONDS = TimeValue.timeValueSeconds(10).millis().toInt()
+
+    val client = createHttpClient()
+
+    /**
+     * Create [CloseableHttpAsyncClient] as a [PrivilegedAction] in order to avoid [java.net.NetPermission] error.
+     */
+    private fun createHttpClient(): CloseableHttpAsyncClient {
+        val config = RequestConfig.custom()
+                .setConnectTimeout(CONNECTION_TIMEOUT_MILLISECONDS)
+                .setConnectionRequestTimeout(REQUEST_TIMEOUT_MILLISECONDS)
+                .setSocketTimeout(SOCKET_TIMEOUT_MILLISECONDS)
+                .build()
+
+        return AccessController.doPrivileged(PrivilegedAction<CloseableHttpAsyncClient>({
+            HttpAsyncClientBuilder.create()
+                    .setDefaultRequestConfig(config)
+                    .useSystemProperties()
+                    .build()
+        } as () -> CloseableHttpAsyncClient))
+    }
+}
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
@@ -14,25 +14,25 @@
  */
 package com.amazon.opendistroforelasticsearch.alerting.resthandler
 
+import com.amazon.opendistroforelasticsearch.alerting.AlertingPlugin
 import com.amazon.opendistroforelasticsearch.alerting.core.ScheduledJobIndices
+import com.amazon.opendistroforelasticsearch.alerting.core.httpapi.toConstructedUrl
+import com.amazon.opendistroforelasticsearch.alerting.core.model.HttpInput
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob.Companion.SCHEDULED_JOBS_INDEX
 import com.amazon.opendistroforelasticsearch.alerting.model.Monitor
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.ALERTING_MAX_MONITORS
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.INDEX_TIMEOUT
-import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.REQUEST_TIMEOUT
-import com.amazon.opendistroforelasticsearch.alerting.util.REFRESH
-import com.amazon.opendistroforelasticsearch.alerting.util._ID
-import com.amazon.opendistroforelasticsearch.alerting.util._VERSION
-import com.amazon.opendistroforelasticsearch.alerting.AlertingPlugin
-import com.amazon.opendistroforelasticsearch.alerting.core.model.HttpInput
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.MAX_ACTION_THROTTLE_VALUE
+import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.REQUEST_TIMEOUT
 import com.amazon.opendistroforelasticsearch.alerting.util.IF_PRIMARY_TERM
 import com.amazon.opendistroforelasticsearch.alerting.util.IF_SEQ_NO
 import com.amazon.opendistroforelasticsearch.alerting.util.IndexUtils
+import com.amazon.opendistroforelasticsearch.alerting.util.REFRESH
+import com.amazon.opendistroforelasticsearch.alerting.util._ID
 import com.amazon.opendistroforelasticsearch.alerting.util._PRIMARY_TERM
 import com.amazon.opendistroforelasticsearch.alerting.util._SEQ_NO
-import org.apache.http.client.utils.URIBuilder
+import com.amazon.opendistroforelasticsearch.alerting.util._VERSION
 import org.apache.logging.log4j.LogManager
 import org.elasticsearch.action.ActionListener
 import org.elasticsearch.action.admin.indices.create.CreateIndexResponse
@@ -46,7 +46,6 @@ import org.elasticsearch.action.support.WriteRequest
 import org.elasticsearch.action.support.master.AcknowledgedResponse
 import org.elasticsearch.client.node.NodeClient
 import org.elasticsearch.cluster.service.ClusterService
-import org.elasticsearch.common.Strings
 import org.elasticsearch.common.settings.Settings
 import org.elasticsearch.common.unit.TimeValue
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler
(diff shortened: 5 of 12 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -27,6 +27,14 @@ Please see our [documentation](https://opendistro.github.io/for-elasticsearch-do
...
 1. Launch Intellij IDEA, choose **Import Project**, and select the `settings.gradle` file in the root of this package. 
 1. To build from the command line, set `JAVA_HOME` to point to a JDK >= 12 before running `./gradlew`.
+  - Unix System
+    1. `export JAVA_HOME=jdk-install-dir`: Replace `jdk-install-dir` by the JAVA_HOME directory of your system.
+    1. `export PATH=$JAVA_HOME/bin:$PATH`
+ 
+  - Windows System
+    1. Find **My Computers** from file directory, right click and select **properties**.
+    1. Select the **Advanced** tab, select **Environment variables**.
+    1. Edit **JAVA_HOME** to path of where JDK softeare is installed.
```
vinooamzn: nitpick: typo. softeare -> software.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -292,7 +298,14 @@ class MonitorRunner(
                         val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
                         results += searchResponse.convertToMap()
                     }
-                    else -> {
+                    is HttpInput -> {
+                        val httpClient = HttpInputClient()
```
vinooamzn: Avoid creating a new client on each input.  Use a single instance.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/httpapi/HttpExtensions.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.elasticsearch.common.xcontent.NamedXContentRegistry
+import org.elasticsearch.common.xcontent.XContentType
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+suspend fun <C : HttpAsyncClient, T> C.suspendUntil(block: C.(FutureCallback<T>) -> Unit): T =
+        suspendCancellableCoroutine { cont ->
+            block(object : FutureCallback<T> {
+                override fun cancelled() {
+                    cont.resumeWith(Result.failure(CancellationException("Request cancelled")))
```
vinooamzn: can we use `cont.resumeWithException(CancellationException(...))` here?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder
+import org.elasticsearch.common.unit.TimeValue
+import java.security.AccessController
+import java.security.PrivilegedAction
+
+/**
+ * This class takes [HttpInput] and performs GET requests to given URIs.
+ */
+class HttpInputClient {
```
vinooamzn: This is duplicating functionality in https://github.com/opendistro-for-elasticsearch/alerting/blob/master/notification/src/main/java/com/amazon/opendistroforelasticsearch/alerting/destination/client/DestinationHttpClient.java#L52 . 

We should use the same client & configuration for both HTTP notifications and inputs.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,141 @@
...
+ * This is a data class of HTTP type of input for Monitors.
+ */
+data class HttpInput(
+    val scheme: String,
+    val host: String?,
+    val port: Int,
+    val path: String?,
+    val params: Map<String, String>,
+    val url: String,
+    val connection_timeout: Int,
```
vinooamzn: https://github.com/opendistro-for-elasticsearch/alerting/blob/38d2c20274cebb856e9cf20cda8c97b826fadbaf/core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt#L41-L42

`connect_timeout` and `socket_timeout` should be part of the client settings not the input.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,141 @@
...
+    init {
+        require(connection_timeout > 0) {
+            "Connection timeout: $connection_timeout is not greater than 0."
+        }
+        require(socket_timeout > 0) {
+            "Socket timeout: $socket_timeout is not greater than 0."
+        }
+
+        // Create an UrlValidator that only accepts "http" and "https" as valid scheme and allows local URLs.
+        val urlValidator = UrlValidator(arrayOf("http", "https"), UrlValidator.ALLOW_LOCAL_URLS)
```
vinooamzn: Seems to me like we shouldn't `ALLOW_LOCAL_URLS`?

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,141 @@
...
+
+        // Build url field by field if not provided as whole, and update url field.
+        val constructedUrl = if (Strings.isNullOrEmpty(url)) {
```
vinooamzn: url can't be null since it's declared `String` not `String?`.  Prefer the kotlin `isEmpty` method to `Strings.isNullOrEmpty` in kotlin code: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-empty.html

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,141 @@
...
+        require(connection_timeout > 0) {
+            "Connection timeout: $connection_timeout is not greater than 0."
+        }
+        require(socket_timeout > 0) {
+            "Socket timeout: $socket_timeout is not greater than 0."
+        }
+
+        // Create an UrlValidator that only accepts "http" and "https" as valid scheme and allows local URLs.
+        val urlValidator = UrlValidator(arrayOf("http", "https"), UrlValidator.ALLOW_LOCAL_URLS)
+
```
vinooamzn: You also need to validate that only one of `url` or `scheme+host+port+path..` is set.

core/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInputTest.kt
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            // Invalid path
+            HttpInput("http", "localhost", 9200, "///", mapOf(), "", 5000, 5000)
+            fail("Invalid path when creating HttpInput should fail.")
+        } catch (e: IllegalArgumentException) {
+            Assert.assertEquals("Invalid url: http://localhost:9200///", e.message)
+        }
+        try {
+            // Invalid url
+            HttpInput("http", "localhost", 9200, "_cluster/health", mapOf(), "¯¯\\_( ͡° ͜ʖ ͡°)_//¯ ", 5000, 5000)
```
vinooamzn: 😂

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: Decided to not to have a single instance of `HttpInputClient` because it needs to be recreated every time when closed, and it would not be good to leave a single instance running the whole time. This is due to the way `HttpAsyncClient` works.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
ann3431: `Notification` is not using asynchronous calls. It would be nice to create a shared parent class for both in the future.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: These can be set from `HttpInput` for now, as XPack also allows to modify timeout values using inputs. Could probably make this available to set in `/_cluster/setting`

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: Need to allow local urls so that monitors can make API calls to local cluster. There are cases of performing local calls in this issue #47 .

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: Will change to `isEmpty()`.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
vinooamzn: Then we should bound these to an upper limit to avoid clients setting too large timeouts.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
vinooamzn: To monitor the local cluster calls can be made through the LoadBalancer URL (AES) or the hostname (for self-managed clusters).  Allowing localhost access means that the monitor can access any other process listening on the same host.  If we want to allow localhost we need to restrict to 9200 (or whatever the `http.port` setting is).

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
vinooamzn: Can you create an issue for that?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
vinooamzn: What do you mean by "running the whole time"? The client should only be closed when the application shuts down - it doesn't consume any resources (other than a few bytes for the memory when there are no requests running).  The example in the QuickStart guide also uses a single client for making multiple requests: https://hc.apache.org/httpcomponents-asyncclient-dev/quickstart.html. 

Creating a client is an expensive operation and should only be done once.

README.md
ann3431: Thanks, will fix the typo.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/httpapi/HttpExtensions.kt
ann3431: Will do so, thanks!

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: Sure, thinking of using 60 seconds as the maximum since the highest frequency monitors can run is every 60s. Do you have any suggestions for the maximum number?

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: No problem, will add a validation to restrict port to be 9200 when host is equal to localhost.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: I can create one client to reuse, and call `execute(request)` for each request. But then I will not be closing the client anywhere in `MonitorRunner`. Do you have any suggestions on where to put the`client.close()`?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
ann3431: Sure! Will do so.

ann3431:
> We also need some limit on response size as over-large responses can take down the cluster.

@vinooamzn There is no out of box function for the client to limit response size. The way I can think of preventing this problem is to restrict the maximum timeout value so that large amount of data cannot be transmitted in limited time. Do you know any good way to limit the content size?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
vinooamzn: Either stop it in `MonitorRunner.doStop` or you can make `HttpInputClient` a `LifecycleComponent`, register it in `MonitoringPlugin.createComponents` and have the ES server start and stop it.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
vinooamzn: `connect_timeout` should be 5 secs and `socket_timeout` can be 60 secs.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: Added `close()` in `MonitorRunner.doStop()` function. Modified the creation and `start()` of client to `doStart()` and `init` respectively.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: `connection_timeout` changed to 5s

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
ann3431: Issues created for the two client classes: #87

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
ann3431: Issue created regarding making the settings from `_cluster/settings` #86

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -292,7 +301,25 @@ class MonitorRunner(
                         val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
                         results += searchResponse.convertToMap()
                     }
-                    else -> {
+                    is HttpInput -> {
+                        val response: HttpResponse = httpClient.client.suspendUntil {
```
vinooamzn: There are a few problems here.  Firstly the read of the content is still blocking as it uses an input stream which blocks waiting for data. Apache Async HttpClient provides APIs for reading the response asynchronously (streaming). See the sample code here: https://hc.apache.org/httpcomponents-asyncclient-dev/examples.html . 

Alternatively to continue to use blocking I/O you need to run this in the I/O dispatcher `Dispatchers.IO`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -292,7 +301,25 @@ class MonitorRunner(
...
+                        } else {
+                            logger.info("Content-Length header does not exist, check estimate size of content.")
+                            val content = response.entity.content
+                            logger.info("Content has estimate number of ${content.available()} bytes.")
+                            content.available()
+                        }
+                        if (contentLength > httpClient.MAX_CONTENT_LENGTH) {
+                            throw Exception("Response content size: $contentLength, is larger than ${httpClient.MAX_CONTENT_LENGTH}.")
+                        }
+                        results += response.toMap()
```
vinooamzn: Does this work if the response is not a JSON response (for e.g. plain text)?

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/httpapi/HttpExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+}
+
+fun HttpInput.toGetRequest(): HttpGet {
+    // Change timeout values to settings specified from input, multiply by 1000 to convert to milliseconds.
+    val requestConfig = RequestConfig.custom()
+            .setConnectTimeout(this.connection_timeout * 1000)
+            .setSocketTimeout(this.socket_timeout * 1000)
+            .build()
+    // If url field is null or empty, construct an url field by field.
+    val constructedUrl = if (Strings.isNullOrEmpty(this.url)) {
```
vinooamzn: nitpick: `String.isNullOrEmpty` -> `url.isEmpty`

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -582,6 +582,116 @@ class MonitorRunnerIT : AlertingRestTestCase() {
...
+        putAlertMappings() // Required as we do not have a create alert API.
+        val trigger = randomTrigger(condition = Script("""
+            return ctx.results[0].status.equals("red")
+        """.trimIndent()))
+        val clusterIndex = randomInt(clusterHosts.size - 1)
+        val input = randomHttpInput(
+                scheme = clusterHosts[clusterIndex].schemeName,
+                host = clusterHosts[clusterIndex].hostName,
+                port = clusterHosts[clusterIndex].port,
+                path = "_cluster/health",
```
vinooamzn: Need a test with a plain text or other form of non-JSON output.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -292,7 +301,25 @@ class MonitorRunner(
...
                         results += searchResponse.convertToMap()
                     }
-                    else -> {
+                    is HttpInput -> {
+                        val response: HttpResponse = httpClient.client.suspendUntil {
+                            httpClient.client.execute(input.toGetRequest(), it) }
+                        // Check response content length is not larger than 100MB
+                        val contentLengthHeader = response.getFirstHeader("Content-Length").value
+                        val contentLength = if (contentLengthHeader != null) {
+                            logger.info("Content length is $contentLengthHeader")
```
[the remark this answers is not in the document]
ann3431: Got it, will change it.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: No, the method I'm using now will throw an IOException if there's unrecognized token. I have a [test](https://github.com/ann3431/alerting/blob/005b5c2e3236f945a5207453ccf31d18bd56044b/alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt#L585) in `MonitorRunnerIT` that uses the _cat/indices API which returns plain text.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/httpapi/HttpExtensions.kt
ann3431: Thanks!

alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
ann3431: Also included in another comment: I have this [test](https://github.com/ann3431/alerting/blob/005b5c2e3236f945a5207453ccf31d18bd56044b/alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt#L585) in `MonitorRunnerIT` that uses the _cat/indices API which returns plain text. Is this sufficient? Thanks!

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
vinooamzn: Why do we require the output to be JSON?  What about XML, CSV, HTML, plain text which are just as common response `Content-Type`s.   If the response is not JSON we should be including it in `results` as a properly escaped strings.

ann3431: We are planning to support more response types such as plain text, Smile, CBOR, YAML. The types you mentioned are also nice to have. But I won't be available to include more types for initial release. I can create an issue for this once this is pushed to master.  Thanks for your suggestion.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -292,7 +301,25 @@ class MonitorRunner(
...
+                            contentLengthHeader.toInt()
+                        } else {
+                            logger.info("Content-Length header does not exist, check estimate size of content.")
+                            val content = response.entity.content
+                            logger.info("Content has estimate number of ${content.available()} bytes.")
+                            content.available()
```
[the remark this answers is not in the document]
ann3431: Decided to set Alert to Error state if response comes without `Content-Length` header. Tried using `LimitedInputStream` but seems like it still requires calling `read()` in order to see if content size exceeds limit. From previous tests I believe responses from Elasticsearch clusters do have `content-length` headers.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: Decided to change to `results += withContext(Dispatchers.IO) {
                            response.toMap()
                        }`

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,160 @@
...
+            for (e in params.entries)
+                uriBuilder.addParameter(e.key, e.value)
+            uriBuilder.build()
+        } else {
+            URIBuilder(url).build()
+        }
+
+        // Make sure that when host is "localhost", only port 9200 is allowed.
+        if (constructedUrl.host == "localhost") {
+            require(constructedUrl.port == 9200) {
```
jinsoor-amzn: We should check the current ES `http.port` setting to validate the port. In order to do this we need the `Settings` object which might be tricky to get in here.

If we cannot get `Settings` object here , one possible alternative I can think is to allow any port number in the Model but we can check during the runtime in `MonitorRunner`.

ann3431: Moved port validation from `HttpInput` to `RestIndexMonitorAction`.  Using the `http.port` from `Settings` to validate.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -105,7 +111,7 @@ class MonitorRunner(
 ) : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
     private val logger = LogManager.getLogger(MonitorRunner::class.java)
-
+    private var httpClient: HttpInputClient
```
dbbaughe: private lateinit var httpClient: HttpInputClient

alerting/build.gradle
```diff
@@ -49,7 +49,10 @@ configurations.all {
         force "commons-logging:commons-logging:${versions.commonslogging}"
         force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
         force "commons-codec:commons-codec:${versions.commonscodec}"
-        
+        force "commons-collections:commons-collections:3.2.2"
+        force "org.apache.httpcomponents:httpcore-nio:4.4.11"
+        force "org.apache.httpcomponents:httpclient:4.5.7"
```
dbbaughe: Do you still need these?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: Fixed

alerting/build.gradle
ann3431: Yes, deleting either of them will result in conflicts.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -180,6 +184,36 @@ class RestIndexMonitorAction(
             }
         }
 
+        /**
+         * This function checks whether the [Monitor] has an [HttpInput] with localhost. If so, make sure the port is same as specified in settings.
+         */
+        private fun validateLocalPort(monitor: Monitor, settingsPort: Int) {
+            if (monitor.inputs.isNotEmpty() && monitor.inputs.single() is HttpInput) {
```
jinsoor-amzn: I know we currently limit the number of inputs to 1 but no need to limit this code to force only one. Just iterate through all the `input` and any input of type `HttpInput` check for the localhost and port is valid.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
ann3431: Initialized in`init()`, so `lateinit` is not needed.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -180,6 +184,36 @@ class RestIndexMonitorAction(
...
+                val httpInput = monitor.inputs.single() as HttpInput
+                // Build url field by field if not provided as whole.
+                val constructedUrl = if (Strings.isEmpty(httpInput.url)) {
```
jinsoor-amzn: there is similar implementation in `HttpExtensions.kt.HttpInput.toGetRequest()` that uses different `isEmpty`. One uses `org.elasticsearch.common.Strings.isEmpty` and other uses `kotlin.String.isEmpty`. Lets be consistent on what use and even better we can extract the logic to `constructedUrl` and share function in both places.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ann3431: Fixed. Changed to go through `monitor.inputs` and do validation if it is a `HttpInput`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
ann3431: Created a shared function `HttpInput.toConstructedUrl()`in `HttpExtensions.kt`

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/resthandler/RestIndexMonitorAction.kt
```diff
@@ -180,6 +183,27 @@ class RestIndexMonitorAction(
             }
         }
 
+        /**
+         * This function checks whether the [Monitor] has an [HttpInput] with localhost. If so, make sure the port is same as specified in settings.
+         */
+        private fun validateLocalPort(monitor: Monitor, settingsPort: Int) {
+            if (monitor.inputs.isNotEmpty()) {
```
jinsoor-amzn: No need to have if-else case on checking the input is empty. the for-loop is enough to handle that. Also please fix the indentation.

ann3431: Done.

PR #97, opened 2019-08-19
Adds maven publish task to notification

dbbaughe:
*Issue #, if available:*

*Description of changes:*
Adds maven publish task to notification

### DIFF
--- a/notification/build.gradle
+++ b/notification/build.gradle
@@ -13,10 +13,11 @@
  *   permissions and limitations under the License.
  */
 
+apply plugin: 'com.github.johnrengelman.shadow'
 apply plugin: 'java'
 apply plugin: 'jacoco'
 apply plugin: 'maven-publish'
-
+apply plugin: 'signing'
 
 dependencies {
     compileOnly "org.elasticsearch:elasticsearch:${es_version}"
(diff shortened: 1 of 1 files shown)


notification/build.gradle
```diff
@@ -15,6 +15,8 @@
 
 apply plugin: 'java'
 apply plugin: 'jacoco'
+apply plugin: 'maven-publish'
```
jinsoor-amzn: basing off this [Job-Scheduler PR](https://github.com/opendistro-for-elasticsearch/job-scheduler/commit/538a5c2e6c642ac5bb8ec6059953c7df862b6846) this change doesn't have `signing`. Do we need to also add this?

notification/build.gradle
```diff
@@ -24,3 +27,78 @@ dependencies {
     testImplementation "org.elasticsearch.test:framework:${es_version}"
     testImplementation "org.easymock:easymock:4.0.1"
 }
+
+shadowJar {
```
zengyan-amazon: this may not work when you have ISM and alerting both depending on notification. I would suggest to repackage at ES plugin level to avoid JarHell

dbbaughe: Tested and relocation/repackaging seems to happen during publish and not build. Alerting/ISM using notification should be fine for now, but we do need a long term solution for our collective plugins dependencies. Thanks for the help!

PR #143, opened 2020-01-08
Adds settings to disable and delete history

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds settings for disabling alert history and deleting old history indices.

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
@@ -249,7 +249,7 @@ class AlertIndices(
         deleteOldHistoryIndices()
     }
 
-    fun rolloverHistoryIndex(): Boolean {
+    private fun rolloverHistoryIndex(): Boolean {
         if (!historyIndexInitialized) {
             return false
         }
@@ -287,9 +287,14 @@ class AlertIndices(
 
             if ((Instant.now().toEpochMilli() - creationTime) > historyRetentionPeriod.millis) {
                 val alias = indexMetaData.aliases.firstOrNull { HISTORY_WRITE_INDEX == it.value.alias }
-                if (alias != null && historyEnabled) {
-                    // If the index has the write alias and history is enabled, don't delete the index
-                    break
+                if (alias != null) {
+                    if (historyEnabled) {
+                        // If the index has the write alias and history is enabled, don't delete the index
+                        break
+                    } else {
+                        // Otherwise reset historyIndexInitialized since index will be deleted
+                        historyIndexInitialized = false
+                    }
                 }
 
                 indicesToDelete.add(indexMetaData.index.name)
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
@@ -134,6 +134,7 @@ class AlertIndices(
 
     override fun onMaster() {
         try {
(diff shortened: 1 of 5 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -57,20 +63,24 @@ import org.elasticsearch.threadpool.ThreadPool
  */
 class AlertIndices(
     settings: Settings,
-    private val client: IndicesAdminClient,
+    private val client: Client,
```
lucaswin-amzn: Why did we change this to normal client? I see you changed all `client` to `client.admin()` was it just to use the suspend until?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -235,12 +260,61 @@ class AlertIndices(
...
     }
+
+    private fun deleteOldHistoryIndices() {
+        val indicesToDelete = mutableListOf<String>()
+
+        val clusterStateRequest = ClusterStateRequest()
+                .clear()
+                .indices(HISTORY_ALL)
+                .metaData(true)
+                .local(true)
```
lucaswin-amzn: doesn't this mean you are only getting local indices rather than cluster wide indices? I don't recall if this job runs on all nodes, or only on a single one.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
qreshi: With the changes to ```AlertIndices``` now also making requests for index metadata when deleting old history indices, it made sense to change this to ```Client```. That way we can use both ```IndicesAdminClient``` and ```ClusterAdminClient``` as necessary within the class, similar to how it is in ISM.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
qreshi: Right, using ```local(true)``` will have it run on master in this case, which is the behavior we want here.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/alerts/AlertIndices.kt
```diff
@@ -235,12 +260,61 @@ class AlertIndices(
...
+
+        val clusterStateResponse = client.admin().cluster().state(clusterStateRequest).actionGet()
+
+        for (entry in clusterStateResponse.state.metaData.indices) {
+            val indexMetaData = entry.value
+            val creationTime = indexMetaData.creationDate
+
+            if ((Instant.now().toEpochMilli() - creationTime) > historyRetentionPeriod.millis) {
+                val alias = indexMetaData.aliases.firstOrNull { HISTORY_WRITE_INDEX == it.value.alias }
+                if (alias != null && historyEnabled) {
```
dbbaughe: If `historyEnabled = false`, it means we'll end up deleting the history index w/ the `HISTORY_WRITE_INDEX` alias. But, the `historyIndexInitialized` could still be cached as true and if the user re-enables history then a new history index w/ the `HISTORY_WRITE_INDEX` will not be created and things will start to fail.

qreshi: Good catch, added logic to reset ```historyIndexInitialized``` in this case to allow the history index to be recreated when ```historyEnabled``` is set back to ```true```

PR #174, opened 2020-03-10
Push Notification Jar to Maven

rishabh6788:
As of now the ISM plugin team has to depend on infra team to push the notification and job-scheduler jars to maven before they start the build process for new release.
This job will push the notification jar to staging repository in maven from where the user can publish the jar into prod. The job will be triggered whenever a new tag is cut for the new release.

### DIFF
--- a/.github/workflows/push-notification-jar.yml
+++ b/.github/workflows/push-notification-jar.yml
@@ -1,0 +1,47 @@
+name: Upload Notification Jar to Maven
+
+on:
+  push:
+    tags: 
+      - v*
+jobs:
+  upload-notification-jar:
+    runs-on: [ubuntu-16.04]
+    strategy:
+      matrix:
+        java: [12]
+    name: Upload Notification Jar to Maven
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
+        
+      - name: Configure AWS CLI
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+
+      - name: Upload Notification Jar to Maven
+        env:
+          passphrase: ${{ secrets.PASSPHRASE }}
+        run: |
+          cd ..
+          export JAVA12_HOME=$JAVA_HOME
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-public-key .
+          aws s3 cp s3://opendistro-docs/github-actions/pgp-private-key .
+          
+          gpg --import pgp-public-key
+          gpg --allow-secret-key-import --import pgp-private-key
+
+          mkdir /home/runner/.gradle
+          aws s3 cp s3://opendistro-docs/github-actions/gradle.properties /home/runner/.gradle/
+          
+          cd alerting/notification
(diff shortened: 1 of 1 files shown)


.github/workflows/push-notification-jar.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  upload-notification-jar:
+    runs-on: [ubuntu-16.04]
+    strategy:
+      matrix:
+        java: [12]
+    name: Upload Notification Jar to Maven
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v1
```
dbbaughe: Can we use checkout v2

.github/workflows/push-notification-jar.yml
```diff
@@ -0,0 +1,47 @@
+name: Upload Notification Jar to Maven
+
+on:
+  push:
+    tags: 
+      - v*
+jobs:
+  upload-notification-jar:
+    runs-on: [ubuntu-16.04]
```
dbbaughe: Why specifically 16.04 and not latest?

.github/workflows/push-notification-jar.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
```
dbbaughe: I don't think we are ever going to parallelize runs across different java versions for a specific version of the plugin. Can we remove the java matrix and just set the java-version here? Ideally we could get the java version from some property file so we don't have to update this though.

.github/workflows/push-notification-jar.yml
```diff
@@ -0,0 +1,47 @@
...
+          
+          gpg --import pgp-public-key
+          gpg --allow-secret-key-import --import pgp-private-key
+
+          mkdir /home/runner/.gradle
+          aws s3 cp s3://opendistro-docs/github-actions/gradle.properties /home/runner/.gradle/
+          
+          cd alerting/notification
+          
+          ../gradlew publishShadowPublicationToSonatype-stagingRepository -Dcompiler.java=12 -Dbuild.snapshot=false -Djavax.net.ssl.trustStore=/opt/hostedtoolcache/jdk/12.0.2/x64/lib/security/cacerts
```
dbbaughe: Does this task assemble the jar? Or am I missing where that's happening 🧐

.github/workflows/push-notification-jar.yml
rishabh6788: 160.04 is LTS and Stable.

.github/workflows/push-notification-jar.yml
rishabh6788: I was under the impression that JAVA_HOME path needs to point to Java version 12. 
If that is not the case then I can go ahead and remove it.

.github/workflows/push-notification-jar.yml
rishabh6788: yes, it assembles the jar files and pushes them to staging repository in https://oss.sonatype.org/

.github/workflows/push-notification-jar.yml
rishabh6788: updated

.github/workflows/push-notification-jar.yml
rishabh6788: removed java matrix as per our call

PR #180, opened 2020-03-18
Added release workflow

gaiksaya:
Created a release workflow that is triggered upon creating tags and will upload the artifacts to s3

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -2,9 +2,6 @@ name: Release workflow
 # This workflow is triggered on creating tags to master or a opendistro release branch
 on:
   push:
-    branches:
-      - master
-      - opendistro-*
     tags:
       -'v*'
 
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       # This step uses the checkout Github action: https://github.com/actions/checkout
       - name: Checkout Branch
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
       # This step uses the setup-java Github action: https://github.com/actions/setup-java
       - name: Set Up JDK ${{ matrix.java }}
         uses: actions/setup-java@v1
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -1,0 +1,86 @@
+name: Release workflow
+# This workflow is triggered on creating tags to master or a opendistro release branch
+on:
+  push:
+  tags:
+    -'v*'
+    branches:
+      - master
+      - opendistro-*
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        java: [12]
+    # Job name
+    name: Build Alerting with JDK ${{ matrix.java }}
+    # This job runs on Linux
+    runs-on: ubuntu-latest
+    steps:
+      # This step uses the checkout Github action: https://github.com/actions/checkout
+      - name: Checkout Branch
+        uses: actions/checkout@v1
+      # This step uses the setup-java Github action: https://github.com/actions/setup-java
+      - name: Set Up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+
+      # Building zip, deb and rpm files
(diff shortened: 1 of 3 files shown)


.github/workflows/test-workflow.yml
```diff
@@ -1,6 +1,6 @@
 name: Test Workflow
 # This workflow is triggered on pull requests to master or a opendistro release branch
-on:
+on: repository_dispatch
```
dbbaughe: I think this can be removed?

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,86 @@
+name: Release workflow
+# This workflow is triggered on creating tags to master or a opendistro release branch
+on:
+  push:
```
dbbaughe: Did you come up with which was better between push and create? Can you share any info you read on why choosing one over the other?

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,86 @@
...
+  tags:
+    -'v*'
+    branches:
+      - master
+      - opendistro-*
```
dbbaughe: @qreshi, we'll need to backport this to every other branch for it to be able to run on those right? Or does this only need to be in master, I forget.

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name: Upload Release Asset
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_name: alerting.zip
+          asset_path: alerting.zip
```
dbbaughe: This alerting.zip is a zip of [plugin.zip, plugin.deb, plugin.rpm] right?
Is it possible with this action to instead upload them individually?
Just imagining someone wants to download the plugin.zip and this makes them download all 3 and then requires them to unzip just to get the plugin.zip

.github/workflows/test-workflow.yml
gaiksaya: Yeah I think so since we are not using curl command now

.github/workflows/release-workflow.yml
gaiksaya: I read that some people faced issue with push but then the security plugin is tried and tested with this one so I guess it should work fine

.github/workflows/release-workflow.yml
qreshi: Yeah, we had to backport our test workflow to older ```opendistro-*``` branches for the jobs to get picked up during PRs, I'm sure it'll be the same case here.

.github/workflows/release-workflow.yml
gaiksaya: As per my research and I consulted Sujith too, for each package (zip, rpm and deb) we need to create upload-release-asset action separately.Basically repeat the action for each file individually

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,110 @@
...
+      matrix:
+        java: [12]
+    # Job name
+    name: Build Alerting with JDK ${{ matrix.java }}
+    # This job runs on Linux
+    runs-on: ubuntu-latest
+    steps:
+      # This step uses the checkout Github action: https://github.com/actions/checkout
+      - name: Checkout Branch
+        uses: actions/checkout@v1
```
qreshi: Might be a good idea to change this to ```checkout@v2```. There is a known issue where reruns using this GitHub action [end up failing](https://github.com/actions/checkout/issues/23) and I believe it was fixed in ```v2```.

.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,111 @@
+name: Release workflow
+# This workflow is triggered on creating tags to master or a opendistro release branch
+on:
+  push:
+    branches:
```
dbbaughe: Question about this setup, is this:
1. Triggering this workflow for all push events to the below branches or matching tags
or
2. Triggering this workflow for all push events that match the tags in only these branches

We definitely don't want #1, and concerned that might be the case here.

PR #184, opened 2020-03-25
Adds support for Elasticsearch 7.5.2

dbbaughe:
*Issue #, if available:*

*Description of changes:*
* Updates internal usage of testing frameworks to new testclusters framework
* Updates where jacoco dumps execution data coverage as gradle tasks were removed
* Removes multi node task and replaces with project property that can be used on run or integTest tasks
* Updates tests that were failing multi-node integration tests
* Updates debugging functionality, can debug multi-node integration tests and run task
* Updates README
* Adds task tree utility dependency so I don't have to keep adding it when I want to see the task tree
* Updates httpclient dependency
* Fixes stats API being broken on multi-node cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,19 +42,25 @@ However, to build the `alerting` plugin subproject, we also use the Elastic buil
 
 1. `./gradlew build` builds and tests all subprojects.
 2. `./gradlew :alerting:run` launches a single node cluster with the alerting plugin installed.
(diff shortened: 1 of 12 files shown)


dbbaughe:
Interesting. This workflow was stuck on Expected status and was not running. I created a branch directly on origin alerting and created a PR from that branch which triggered the workflow for the new branch. But, this PR which was from my fork has used that same workflow that ran for the other branch as a passed check.

PR #190, opened 2020-03-27
Changed upload release asset config

gaiksaya:
*Issue #, if available:*

*Description of changes:* Changed upload release asset config and removed the duplication of the release action

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -65,41 +65,17 @@ jobs:
           draft: true
           prerelease: false
 
-      # Upload the release with .zip as asset
+      # Upload the release with .zip,.rpm and .deb as assets
       - name: Upload Release Asset
         id: upload-release-asset
-        uses: actions/upload-release-asset@v1.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-workflow.yml
```diff
@@ -65,41 +65,17 @@ jobs:
...
-          asset_path: alerting-artifacts/*.rpm
-          asset_content_type: application/zip
-
-      # Upload the release with .deb as asset
-      - name: Upload Release Asset
-        id: upload-release-asset
-        uses: actions/upload-release-asset@v1.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: softprops/action-gh-release@v1
```
dbbaughe: It looks like this action will default to using the tag as the release name for the name input:
https://github.com/softprops/action-gh-release#inputs

But, in the action previous to this one, we have `release_name: Release ${{ env.TAG_VERSION }}`, but there is also a `tag_name`.
Will this action be able to correctly find the release to upload the artifacts?

gaiksaya: I guess it should work if I add the following:

if: startsWith(github.ref, 'refs/tags/')
It will find the matching tag and then release accordingly

PR #194, opened 2020-03-31
Modified upload asset path

gaiksaya:
Updated and tested the workflow especially upload release asset action [test link] (https://github.com/gaiksaya/alerting/runs/547542339?check_suite_focus=true)

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -5,7 +5,6 @@ on:
     tags:
       - 'v*'
 
-
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-workflow.yml
```diff
@@ -29,15 +28,19 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew build buildDeb buildRpm --no-daemon -Dbuild.snapshot=false
 
+      # Creating artifact path as well as individual folders for rpm, zip and deb assets
       - name: Create Artifact Path
         run: |
           mkdir -p alerting-artifacts
```
dbbaughe: Looks like you create `alerting-artifacts` directory before trying to copy the artifacts into it, whereas you're directly copying the artifacts into the new directories before creating them, will this cause any issues?

.github/workflows/release-workflow.yml
```diff
@@ -73,7 +76,7 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_name: alerting.zip
-          asset_path: alerting-artifacts/*.zip
+          asset_path: alerting-artifacts_zip
```
dbbaughe: Will this asset_path work on a directory?

.github/workflows/release-workflow.yml
gaiksaya: For each build, I assumed there would be one each zip, rpm and deb files. But if there is a possibility for more then I need to create directories for each one before copying them. Right now it is basically renaming the zip, rpm and deb file and using that to attach as asset.

PR #220, opened 2020-07-02
Add a new input type for monitors - HttpInput (Inherited from PR #82)

ghost:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/alerting/issues/47

*Description of changes:*
This PR is inherited from https://github.com/opendistro-for-elasticsearch/alerting/pull/82,and here are the additional changes. I have also marked them in the code.
 - Re-base the code from opendistro-1.1 branch (ES 7.1.1) to opendistro-1.9 branch (ES 7.8.0)
 - Change the parameter name from `params` to `query_params` in the JSON string builder in funtion `HttpInput:toXContent()` to be able to reuse `URLInfo` component for the custom webhook URL in Kibana plugin
 - Change the URL path in the unit test `test monitor HttpInput with non JSON response ` due to the hidden index change
 - In `alerting/build.gradle`, add comments for the new configuration of the dependency versions, and remove `httpasyncclient` from dependency as it has been declared in `alerting-core` subproject

As the original PR has been merged (but reverted later on: https://github.com/opendistro-for-elasticsearch/alerting/pull/161, https://github.com/opendistro-for-elasticsearch/alerting/pull/162), I assume all the changes are reviewed, so I changed the code as less as possible.

### DIFF
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -81,7 +81,7 @@ dependencies {
 }
 
 javadoc.enabled = false // turn off javadoc as it barfs on Kotlin code
-licenseHeaders.enabled = true
+licenseHeaders.enabled = false
 dependencyLicenses.enabled = false
 // no need to validate pom, as this project is not uploaded to sonatype
 validateNebulaPom.enabled = false
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -59,7 +59,10 @@ configurations.all {
         force "commons-logging:commons-logging:${versions.commonslogging}"
         force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
         force "commons-codec:commons-codec:${versions.commonscodec}"
-        
+        force "commons-collections:commons-collections:3.2.2"
+        force "org.apache.httpcomponents:httpcore-nio:4.4.11"
+        force "org.apache.httpcomponents:httpclient:4.5.7"
+
         // This is required because kotlin coroutines core-1.1.1 still requires kotin stdlib 1.3.20 and we're using 1.3.21
         force "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
         force "org.jetbrains.kotlin:kotlin-stdlib-common:${kotlin_version}"
@@ -68,6 +71,7 @@ configurations.all {
 
 dependencies {
     compileOnly "org.elasticsearch.plugin:elasticsearch-scripting-painless-spi:${versions.elasticsearch}"
+    compile "org.apache.httpcomponents:httpasyncclient:4.1.4"
 
     // Elasticsearch Nanny state
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -81,7 +81,7 @@ dependencies {
 }
 
 javadoc.enabled = false // turn off javadoc as it barfs on Kotlin code
-licenseHeaders.enabled = false
+licenseHeaders.enabled = true
 dependencyLicenses.enabled = false
 // no need to validate pom, as this project is not uploaded to sonatype
 validateNebulaPom.enabled = false
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -59,9 +59,12 @@ configurations.all {
         force "commons-logging:commons-logging:${versions.commonslogging}"
         force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
         force "commons-codec:commons-codec:${versions.commonscodec}"
+        // commons-validator 1.6 depends on commons-collections 3.2.1 which causes the conflict
         force "commons-collections:commons-collections:3.2.2"
-        force "org.apache.httpcomponents:httpcore-nio:4.4.11"
-        force "org.apache.httpcomponents:httpclient:4.5.7"
+        // httpasyncclient 4.1.4 depends on httpcore-nio 4.4.10 which causes the conflict
+        force "org.apache.httpcomponents:httpcore-nio:4.4.12"
+        // httpasyncclient 4.1.4 depends on httpclient 4.5.6 which causes the conflict
+        force "org.apache.httpcomponents:httpclient:${versions.httpclient}"
 
         // This is required because kotlin coroutines core-1.1.1 still requires kotin stdlib 1.3.20 and we're using 1.3.21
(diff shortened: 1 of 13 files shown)


alerting/src/test/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunnerIT.kt
```diff
@@ -582,6 +582,116 @@ class MonitorRunnerIT : AlertingRestTestCase() {
                 actionResults2[actionThrottleEnabled.id]!!.lastExecutionTime)
     }
 
+    fun `test monitor HttpInput with non JSON response `() {
+        val clusterIndex = randomInt(clusterHosts.size - 1)
+        val input = randomHttpInput(
+                scheme = clusterHosts[clusterIndex].schemeName,
+                host = clusterHosts[clusterIndex].hostName,
+                port = clusterHosts[clusterIndex].port,
+                path = "_cat/plugins",
```
ftianli-amzn, line 591: I changed the path, and the original one is https://github.com/opendistro-for-elasticsearch/alerting/pull/82/files#diff-6a0bc20035e8972567e050ee46d79e13R646

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,153 @@
...
+    override fun name(): String {
+        return HTTP_FIELD
+    }
+
+    companion object {
+        const val SCHEME_FIELD = "scheme"
+        const val HOST_FIELD = "host"
+        const val PORT_FIELD = "port"
+        const val PATH_FIELD = "path"
+        const val PARAMS_FIELD = "query_params"
```
ftianli-amzn, line 102: I changed the name of `PARAMS_FIELD` from "params" to "query_params". (Link for the original code is https://github.com/opendistro-for-elasticsearch/alerting/pull/82/files#diff-5906d6d1ebd2a27e25708de50cfc18cdR102)

alerting/build.gradle
```diff
@@ -59,7 +59,13 @@ configurations.all {
         force "commons-logging:commons-logging:${versions.commonslogging}"
         force "org.apache.httpcomponents:httpcore:${versions.httpcore}"
         force "commons-codec:commons-codec:${versions.commonscodec}"
-        
+        // commons-validator 1.6 depends on commons-collections 3.2.1 which causes the conflict
+        force "commons-collections:commons-collections:3.2.2"
+        // httpasyncclient 4.1.4 depends on httpcore-nio 4.4.10 which causes the conflict
+        force "org.apache.httpcomponents:httpcore-nio:4.4.12"
+        // httpasyncclient 4.1.4 depends on httpclient 4.5.6 which causes the conflict
+        force "org.apache.httpcomponents:httpclient:${versions.httpclient}"
```
ftianli-amzn, line 67: I changed the version number of some dependencies and add some comments ( LInk for the original code is https://github.com/opendistro-for-elasticsearch/alerting/pull/82/files#diff-23c789e267bceb21b7282a6c2e23303fR52 )

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/client/HttpInputClient.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.elasticsearch.common.unit.TimeValue
+import java.security.AccessController
+import java.security.PrivilegedAction
+
+/**
+ * This class takes [HttpInput] and performs GET requests to given URIs.
+ */
+class HttpInputClient {
+
+    // TODO: If possible, these settings should be implemented as changeable via the "_cluster/settings" API.
```
qreshi, line 32: If we wish to do this, these settings can be added to [AlertingSettings](https://github.com/opendistro-for-elasticsearch/alerting/blob/master/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/AlertingSettings.kt). If it isn't considered a high priority right now, we can create an issue to track the change.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,153 @@
...
+    val params: Map<String, String>,
+    val url: String,
+    val connection_timeout: Int,
+    val socket_timeout: Int
+) : Input {
+
+    // Verify parameters are valid during creation
+    init {
+        require(validateFields()) {
+            "Either one of url or scheme + host + port+ + path + params can be set."
```
qreshi, line 47: NP: Redundant `+` between `port` and `path`.

Also can change `can` to `must` in the message since one of these is required.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/HttpInput.kt
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+            return HttpInput(scheme, host, port, path, params, url, connectionTimeout, socketTimeout)
+        }
+    }
+
+    /**
+     * Helper function to check whether one of url or scheme+host+port+path+params is defined.
+     */
+    private fun validateFields(): Boolean {
```
qreshi, line 147: Going off of its name, it look like `validateFields()` should return `true` when the fields are valid. In this case it might be more accurate to call this `areFieldsValid()`.

Also, I might be mistaken but it looks the boolean checks are inverted. `if (url.isNotEmpty())` will only check the `host/port/path/params` when the `url` is not empty, where it should only be checking those values if a `url` is not given. The current logic will return `true` by default when `url` is empty which does not validate the `host/port/path/params`.

Perhaps something like this would make more sense:
```
if (url.isEmpty()) {
    return !(host.isEmpty() && (port == -1) && path.isEmpty() && params.isEmpty())
}

return true
```

ftianli-amzn: Thanks @qreshi for giving the detailed comments! 👍 I will look at the codes.

tlfeng:
Re-opened in PR https://github.com/opendistro-for-elasticsearch/alerting/pull/347

Issue #232, opened 2020-08-06
Associate actions to alerting REST API


skkosuri-amzn:
Currently, Alerting supports 10+ REST API. There API are not associated with actions. This feature is to add actions to the REST API without any changes to REST API behavior. This is similar to Elasticsearch API actions which extends org.elasticsearch.action.ActionType. Actions could be used to apply security permissions to Alerting REST API. 

Proposed actions: 
```
POST    _alerting/destinations - cluster:admin/alerting/destinations/write
PUT    _alerting/destinations/{id} - cluster:admin/alerting/destinations/write
DEL    _alerting/destinations/{id}  - cluster:admin/alerting/destinations/delete 
POST    _alerting/monitors - cluster:admin/alerting/monitors/write
PUT    _alerting/monitors/{id} - cluster:admin/alerting/monitors/write
GET    _alerting/monitors/{id} - cluster:admin/alerting/monitors/read
DEL    _alerting/monitors/{id}  - cluster:admin/alerting/monitors/delete
GET    _alerting/monitors/_search - cluster:admin/alerting/monitors/search
POST    _alerting/monitors/_execute - cluster:admin/alerting/monitors/execute
POST    _alerting/monitors/{id}/_acknowledge/alerts   - cluster:admin/alerting/alerts/ack
```

skkosuri-amzn:
This is completed for all REST API's. All new API's MUST have actions associated with them.

PR #233, opened 2020-08-07
Add action to DELETE /_alerting/destinations/{id}

skkosuri-amzn:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/alerting/issues/232

*Description of changes:*
Refer: https://github.com/opendistro-for-elasticsearch/alerting/issues/232


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/action/DeleteDestinationAction.kt
```diff
@@ -0,0 +1,27 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.alerting.action
+
+import org.elasticsearch.action.ActionType
+import org.elasticsearch.action.delete.DeleteResponse
+import org.elasticsearch.common.io.stream.Writeable
+
+class DeleteDestinationAction private constructor() : ActionType<DeleteResponse>(NAME, Writeable.Reader({ DeleteResponse(null) })) {
```
qreshi, line 22: If we wish to make `DeleteDestinationAction` itself a singleton, is it possible to define this as an `object`, instead of using a `companion object` to store `INSTANCE`?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportDeleteDestinationAction.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.elasticsearch.action.support.HandledTransportAction
+import org.elasticsearch.client.Client
+import org.elasticsearch.common.inject.Inject
+import org.elasticsearch.common.io.stream.Writeable
+import org.elasticsearch.tasks.Task
+import org.elasticsearch.transport.TransportService
+
+class TransportDeleteDestinationAction @Inject constructor(
+    transportService: TransportService,
+    var client: Client,
```
qreshi, line 34: NP: If we don't plan on mutating `client` after assignment we can change it to `val`

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/action/DeleteDestinationAction.kt
skkosuri-amzn: Thanks. Discussed with Drew also, ISM currently using companion object. Keeping as it is for consistency.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportDeleteDestinationAction.kt
skkosuri-amzn: Fixed.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/action/DeleteDestinationRequest.kt
```diff
@@ -22,19 +22,19 @@ import org.elasticsearch.common.io.stream.StreamInput
 import org.elasticsearch.common.io.stream.StreamOutput
 import java.io.IOException
 
-class DeleteDestinationRequest() : ActionRequest() {
+class DeleteDestinationRequest : ActionRequest {
 
-    var destinationID: String? = null
-    var refreshPolicy: WriteRequest.RefreshPolicy? = null
+    val destinationId: String
```
qreshi, line 27: NP: I believe the old naming scheme was more appropriate since we typically only [begin lowercasing acronyms when they exceed two letters](https://kotlinlang.org/docs/reference/coding-conventions.html#choosing-good-names).

This is not a blocker for this PR, we can always change it in another one of the API PRs.

PR #263, opened 2020-10-02
Add AllowList for Destinations

qreshi:
*Issue #, if available:*

*Description of changes:*
Adds AllowList functionality for Destinations
By default all Destination types are enabled

When a Destination is disallowed, referring that type in the Monitor and Destination index APIs will result in a 403 Forbidden error. In the case of the Email Destination, all Email API will throw 403 Forbidden errors when it is disallowed.

If a Destination type that is already in an existing Monitor is removed from the AllowList, the next execution of that Monitor will fail if it tries to publish using that Destination.


skkosuri-amzn (dismissed):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportIndexDestinationAction.kt
```diff
@@ -88,6 +91,19 @@ class TransportIndexDestinationAction @Inject constructor(
         }
 
         private fun prepareDestinationIndexing() {
+
+            // Prevent indexing if the Destination type is disallowed
+            val destinationType = request.destination.type.value
+            if (!allowList.contains(destinationType)) {
+                actionListener.onFailure(
+                    ElasticsearchStatusException(
```
lezzago, line 99: Why is there no `AlertingException.wrap()`?

qreshi: Added.

lezzago (approved):

skkosuri-amzn (approved):

PR #275, opened 2020-10-19
add user role filter for monitor on top of ad result

ylwu-amzn:
*Issue #https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/195*

## Description of changes:

AD result index will be system indices, so create AD monitor will fail as common users have no read permission on anomaly result index.

And AD and Alerting will provide document level access control with user backend role user. This PR adds detector backend role checking when create and run monitor.

**Backend role checking logic for AD monitor creation:**
1.If monitor's user is null, then check if there is detector with null user. For old monitor and detector, they have no user field. With this we can provide backward compatibility. 
2.If monitor's user is not null, but has no backend roles, then will check if there is detector with empty backend role. 
3.If monitor's user has backend roles, will check if there is detector which backend role has any of the monitor's user backend roles.

The same logic for AD monitor's schedule job run, just change detector existence checking to user role filter query.

**Tradeoff**
We have no way to verify if user has AD read permission in alerting plugin now. So we have to check user's backend role no matter the backend role filter open or not. If we don't add backend role checking when the filter setting disabled, that will be a security leak: if user has permission to create monitor, they can get any detector's results even they have no AD read permission. We make a balance on security and user experience and choose to prioritize security.  In next phase, when secure rest client ready or security plugin provides permission check action, we will add AD read permission check and provide complete user experience.

## Test
Add UT and IT cases for AD monitor creation and run.
Have tested by running `./gradlew :alerting:run`.

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -22,6 +22,7 @@ import com.amazon.opendistroforelasticsearch.alerting.core.JobRunner
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
 import com.amazon.opendistroforelasticsearch.alerting.core.model.SearchInput
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.InjectorContextElement
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.addUserRolesFilter
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.convertToMap
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.firstFailureOrNull
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.retry
@@ -323,7 +324,19 @@ class MonitorRunner(
                         XContentType.JSON.xContent().createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, searchSource).use {
                             searchRequest.source(SearchSourceBuilder.fromXContent(it))
                         }
-                        val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
+
+                        lateinit var searchResponse: SearchResponse
+                        // Add user role filter for AD result
+                        if (input.indices.size == 1 && input.indices[0] == ".opendistro-anomaly-results*") {
+                            client.threadPool().threadContext.stashContext().use {
+                                if (monitor.user != null && !monitor.user.backendRoles.isNullOrEmpty()) {
+                                    addUserRolesFilter(monitor.user.backendRoles, searchRequest.source(), "user.backend_roles")
+                                }
+                                searchResponse = client.suspendUntil { client.search(searchRequest, it) }
+                            }
+                        } else {
+                            searchResponse = client.suspendUntil { client.search(searchRequest, it) }
+                        }
                         results += searchResponse.convertToMap()
                     }
                     else -> {
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -22,7 +22,6 @@ import com.amazon.opendistroforelasticsearch.alerting.core.JobRunner
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
 import com.amazon.opendistroforelasticsearch.alerting.core.model.SearchInput
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.InjectorContextElement
-import com.amazon.opendistroforelasticsearch.alerting.elasticapi.addUserRolesFilter
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.convertToMap
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.firstFailureOrNull
 import com.amazon.opendistroforelasticsearch.alerting.elasticapi.retry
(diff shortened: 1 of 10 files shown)


saratvemulapalli (approved):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -323,7 +324,19 @@ class MonitorRunner(
                         XContentType.JSON.xContent().createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, searchSource).use {
                             searchRequest.source(SearchSourceBuilder.fromXContent(it))
                         }
-                        val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
```
skkosuri-amzn, line 326: Roles inject happens here: https://github.com/opendistro-for-elasticsearch/alerting/blob/master/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt#L227
This takes care of keeping context intack when co-routines switch.

ylwu-amzn: Thanks. I add a separate method to collect anomaly result input.

saratvemulapalli (commented):
Changes look good to me.
Will wait for the tests.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -142,6 +149,41 @@ class TransportIndexMonitorAction @Inject constructor(
...
+     * As anomaly detector index is system index, common user has no permission to query. So we need
+     * to send REST API call to AD REST API.
+     */
+    fun checkAnomalyDetectorAndExecute(
+        client: Client,
+        actionListener: ActionListener<IndexMonitorResponse>,
+        request: IndexMonitorRequest
+    ) {
+        client.threadPool().threadContext.stashContext().use {
+            val searchSourceBuilder = SearchSourceBuilder().size(0)
```
lezzago, line 165: This will return 0 results since the size is 0

lezzago, line 165: This will return 0 results everytime

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/AnomalyDetectionUtils.kt
```diff
@@ -0,0 +1,77 @@
...
+        val userExistsQuery = QueryBuilders.existsQuery(userFieldName)
+        val userExistsNestedQueryBuilder = NestedQueryBuilder(userFieldName, userExistsQuery, ScoreMode.None)
+
+        boolQueryBuilder.mustNot(nestedQueryBuilder)
+        boolQueryBuilder.must(userExistsNestedQueryBuilder)
+    } else {
+        // For normal case, user should have backend roles.
+        val userRolesFilterQuery = QueryBuilders.termsQuery(userBackendRoleFieldName, user.backendRoles)
+        val nestedQueryBuilder = NestedQueryBuilder(userFieldName, userRolesFilterQuery, ScoreMode.None)
+        boolQueryBuilder.must(nestedQueryBuilder)
```
lezzago, line 68: These 2 lines are repeated in all of the different `if else` cases

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/elasticapi/ElasticExtensions.kt
```diff
@@ -145,11 +145,14 @@ fun XContentBuilder.optionalUserField(name: String, user: User?): XContentBuilde
...
-    val queryBuilder = searchSourceBuilder.query() as BoolQueryBuilder
-    searchSourceBuilder.query(queryBuilder.filter(filterBckendRoles))
+    addUserRolesFilter(user.backendRoles, searchSourceBuilder, fieldName)
 }
 
+fun addUserRolesFilter(userRoles: List<String>, searchSourceBuilder: SearchSourceBuilder, fieldName: String) {
+    val filterBackendRoles = QueryBuilders.termsQuery(fieldName, userRoles)
+    val queryBuilder = searchSourceBuilder.query() as BoolQueryBuilder
+    searchSourceBuilder.query(queryBuilder.filter(filterBackendRoles))
+}
```
lezzago, line 155: Why do u need to create a new function? If it is because of the function name, I would then rather rename.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportIndexMonitorAction.kt
ylwu-amzn: Yeah, this is by design. We don't need to get the detail content of detector, just need the total hits. Check this line
```
val totalHits = response?.hits?.totalHits?.value
```

ylwu-amzn: duplicate comments, resolve this one

lezzago: ah ok

saratvemulapalli (commented):
Changes overall look good to me. 
Thanks for the changes.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/AnomalyDetectionUtils.kt
ylwu-amzn: Not all branches. line 68 and line 63 looks like duplicate, but if I remove one line of them, I need to add another 2 lines: 
Sample code
```
var mustQuery = null
if (...) {
   mustQuery = userExistsNestedQueryBuilder
} else {
   mustQuery = nestedQueryBuilder
}
boolQueryBuilder.must(mustQuery)
```
Prefer to keep as is.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/elasticapi/ElasticExtensions.kt
ylwu-amzn: This change is from my previous commit, in that commit, I need this function. Will remove this one to make the change minimal .

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/AnomalyDetectionUtils.kt
lezzago: ah ok, didnt realize that. Thanks

lezzago (approved):

skkosuri-amzn (approved):

PR #282, opened 2020-10-24
resolve user first before add user role filter for AD

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
Request from REST API doesn't contain auth `User` object. Will get user from security plugin first, then use it to build user role filter for AD


skkosuri-amzn (approved):

lezzago (approved):

skkosuri-amzn:
thanks @ylwu-amzn

PR #312, opened 2020-12-01
Run /_execute in User context

skkosuri-amzn:
*Issue #, if available:*
Run /_execute in User context

*Description of changes:*
Run /_execute in User context

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
@@ -7,6 +7,8 @@ import com.amazon.opendistroforelasticsearch.alerting.action.ExecuteMonitorRespo
 import com.amazon.opendistroforelasticsearch.alerting.core.model.ScheduledJob
 import com.amazon.opendistroforelasticsearch.alerting.model.Monitor
 import com.amazon.opendistroforelasticsearch.alerting.util.AlertingException
+import com.amazon.opendistroforelasticsearch.commons.ConfigConstants
+import com.amazon.opendistroforelasticsearch.commons.authuser.User
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -39,8 +41,14 @@ class TransportExecuteMonitorAction @Inject constructor(
 ) : HandledTransportAction<ExecuteMonitorRequest, ExecuteMonitorResponse> (
         ExecuteMonitorAction.NAME, transportService, actionFilters, ::ExecuteMonitorRequest) {
 
+    private var user: User? = null
+
     override fun doExecute(task: Task, execMonitorRequest: ExecuteMonitorRequest, actionListener: ActionListener<ExecuteMonitorResponse>) {
 
+        val userStr = client.threadPool().threadContext.getTransient<String>(ConfigConstants.OPENDISTRO_SECURITY_USER_AND_ROLES)
+        log.debug("User and roles string from thread context: $userStr")
```
qreshi, line 49: I don't remember if `log.debug()` prepends any context information but if it doesn't, could we make this message a little more specific to know that it's coming from the execute transport action?

skkosuri-amzn: Double checked, it looks like below: `[2020-12-01T00:10:12,201][DEBUG][c.a.o.a.t.TransportIndexMonitorAction] [38f9d35bb887.ant.amazon.com] User and roles string from thread context: admin|admin|own_index,all_access`

It has the transport class name.

lezzago (approved):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
qreshi: Cool, thanks for checking

qreshi (approved):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -87,7 +95,10 @@ class TransportExecuteMonitorAction @Inject constructor(
                     }
                 })
             } else {
-                val monitor = execMonitorRequest.monitor as Monitor
+                val monitor = when (user == null || user?.name.isNullOrEmpty()) {
+                    true -> execMonitorRequest.monitor as Monitor
+                    false -> (execMonitorRequest.monitor as Monitor).copy(user = user)
```
ylwu-amzn, line 100: If user has permission to run mointor, then they can run other user's monitor?

skkosuri-amzn: The flow is same, if User roles has permission to run monitor and read permission for the index being monitored, yes.

ylwu-amzn: If user enabled user backend role filter for alerting, then they should not see/run other user's monitor? Even user has read permission for the index being monitored and run monitor permission, user should not be able to see other user's monitor on Kibana, so they can't execute other user's monitor from Kibana. But if they know other user's monitor id, they can execute the monitor?
What problem will this PR address ? Per my understanding, this PR will fix such case: user in monitor has no run monitor permission, but the person who execute the monitor has run monitor permission, for such case, this PR will allow that person to run monitor successfully.

PR #328, opened 2020-12-22
Add check to Destination to block reserved IP addresses

ghost:
*Issue #, if available:*

*Description of changes:*
**Woring In Progress**
 - Block reserved IP addresses from the list: https://en.wikipedia.org/wiki/Reserved_IP_addresses to be used as the notification destinations of alerts.


ftianli-amzn:
The unit test of the newly added codes in `DestinationHttpClient.java` was covered in `CustomWebhookMessageTest.java`, so the code coverage is not decreased actually.

PR #335, opened 2021-01-11
Move to new ktlint version

skkosuri-amzn:
*Issue #, if available:*
Move to new ktlint version.

*Description of changes:*
Move to new ktlint version.

### DIFF
--- a/.editorConfig
+++ b/.editorConfig
@@ -2,3 +2,4 @@ root=true
 [*.{kt,kts}]
 insert_final_newline=true
 max_line_length=140
+disabled_rules=import-ordering
--- a/build.gradle
+++ b/build.gradle
@@ -71,19 +71,19 @@ configurations {
 }
 
 dependencies {
-    ktlint "com.github.shyiko:ktlint:0.29.0"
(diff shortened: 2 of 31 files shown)


.editorConfig
```diff
@@ -2,3 +2,4 @@ root=true
 [*.{kt,kts}]
 insert_final_newline=true
 max_line_length=140
+disabled_rules=import-ordering
```
ftianli-amzn, line 5: I understand the rule of `import-ordering` is probably nonsensical and may cause problem (https://github.com/pinterest/ktlint/issues/ 527), but looks like it doesn't work as expected according to the result the `Test Workflow`
There are still errors in the log: `Imports must be ordered in lexicographic order without any empty lines in-between`.

build.gradle
```diff
@@ -71,19 +71,19 @@ configurations {
 }
 
 dependencies {
-    ktlint "com.github.shyiko:ktlint:0.29.0"
+    ktlint "com.pinterest:ktlint:0.34.2"
```
ftianli-amzn, line 74: Just curious, the version `0.34.2` was released in 2019, and do we plan to use a newer version?

Issue #91, opened 2019-08-08
Add Amazon SNS destination in Alerting


monitoringit:
Is there a way to build Amazon SNS destination in opendistro Alerting?

Amazon SNS destination is available in Aws elasticsearch service but when we install opendistro plugin in self hosted elasticsearch i only see amazon chime, custom webhook and slack as the only options available? I am using 6.7.1 elasticsearch.

dbbaughe:
Hi @monitoringit,

Currently we do not support Amazon SNS destination in Open Distro version. We need to create an Open Distro friendly implementation of Amazon SNS destination.

We can keep this open to track progress and future requests of this destination.

Thanks,
Drew

monitoringit:
Thanks Drew!
Is that something i can build custom into my stack? I really like amazon
sns and would love to add it in opendistro as destination. Any pointers
will be helpful.

Also, can custom webhook to be used to send messages to amazon sns? If yes,
do you have any snippets of that config?

qreshi:
Closing in favor of https://github.com/opensearch-project/alerting/issues/45

PR #348, opened 2021-02-08
Support SNS as a destination

lezzago:
*Issue #, if available:*
#91 
*Description of changes:*
Support SNS as a destination

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -142,6 +142,9 @@ class MonitorRunner(
     fun reloadDestinationSettings(settings: Settings) {
         destinationSettings = loadDestinationSettings(settings)
 
+        // Update awsSettings for SNS destination type
+        awsSettings = AWSSettings.parse(settings)
+
         // Update destinationContextFactory as well since destinationSettings has been updated
         destinationContextFactory.updateDestinationSettings(destinationSettings)
     }
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
@@ -47,6 +47,7 @@ import com.amazon.opendistroforelasticsearch.alerting.model.action.Action.Compan
 import com.amazon.opendistroforelasticsearch.alerting.model.destination.DestinationContextFactory
 import com.amazon.opendistroforelasticsearch.alerting.script.TriggerExecutionContext
 import com.amazon.opendistroforelasticsearch.alerting.script.TriggerScript
+import com.amazon.opendistroforelasticsearch.alerting.settings.AWSSettings
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.ALERT_BACKOFF_COUNT
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.ALERT_BACKOFF_MILLIS
 import com.amazon.opendistroforelasticsearch.alerting.settings.AlertingSettings.Companion.MOVE_ALERTS_BACKOFF_COUNT
@@ -123,6 +124,7 @@ class MonitorRunner(
 
     @Volatile private var destinationSettings = loadDestinationSettings(settings)
     @Volatile private var destinationContextFactory = DestinationContextFactory(client, xContentRegistry, destinationSettings)
+    @Volatile private var awsSettings = AWSSettings.parse(settings)
 
     init {
         clusterService.clusterSettings.addSettingsUpdateConsumer(ALERT_BACKOFF_MILLIS, ALERT_BACKOFF_COUNT) {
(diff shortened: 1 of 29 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/destination/Destination.kt
```diff
@@ -229,14 +240,15 @@ data class Destination(
...
+                sns = SNS.readFrom(sin),
                 customWebhook = CustomWebhook.readFrom(sin),
                 email = Email.readFrom(sin)
             )
         }
     }
 
     @Throws(IOException::class)
-    fun publish(compiledSubject: String?, compiledMessage: String, destinationCtx: DestinationContext): String {
+    fun publish(AWSSettings: AWSSettings, compiledSubject: String?, compiledMessage: String, destinationCtx: DestinationContext): String {
```
qreshi, line 251: camelCase parameter name to `awsSettings`

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -123,6 +124,7 @@ class MonitorRunner(
 
     @Volatile private var destinationSettings = loadDestinationSettings(settings)
     @Volatile private var destinationContextFactory = DestinationContextFactory(client, xContentRegistry, destinationSettings)
+    @Volatile private var awsSettings = AWSSettings.parse(settings)
```
qreshi, line 127: The way this is currently, the `access/secret` key settings will be loaded from the keystore only upon bootstrapping of ES. If the user wanted to change those settings, they would need to restart the cluster which isn't ideal.

You might want to extend `reloadDestinationSettings` to support updating these settings using the reload secure settings API (`POST _nodes/reload_secure_settings`)

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/DestinationType.kt
```diff
@@ -18,11 +18,16 @@ package com.amazon.opendistroforelasticsearch.alerting.util
...
     CUSTOM_WEBHOOK("custom_webhook"),
     EMAIL("email"),
     TEST_ACTION("test_action");
 
     override fun toString(): String {
         return value
     }
+
+    companion object {
+        var snsUseIamRole = false
```
qreshi, line 31: This seems a little strange to me. I'm not sure if we should maintain state on using IAM role in the `DestinationType` enum class.

It looks like this is mostly used for null assertion in the `SNS` data class. Perhaps you can remove this variable and instead validate if either `roleARN` or `access/secret` is set somewhere. It looks like you something like that in `SNSMessage` already, that might be sufficient.

And ideally, we should allow the `access/secret` settings to be changed using the reload API and if we support that then you can have scenarios where:

1. `access/secret` settings are originally set
2. SNS Destinations are created (allowing `roleARN` to be null since `snsUseIamRole = false`)
3. The `access/secret` settings are removed, checking `roleARN` on any subsequent SNS Destination but now you have one that's already created with a null `roleARN` and no `access/secret` settings so it bypassed this check.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/destination/Destination.kt
lezzago: Fixed

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
lezzago: thanks for pointing that out, fixed

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/model/destination/Destination.kt
lezzago: Fixed

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/DestinationType.kt
lezzago: I need this in case the customer uses the indexDestination API to create the SNS destination and we need some validation for the role arn.
Also I have updated the code such that `snsUseIamRole` can change whenever the settings are reloaded.

skkosuri-amzn (approved):

lezzago:
There are some problems discovered, will reopen PR once they have been resolved

alerting/src/main/plugin-metadata/plugin-security.policy
```diff
@@ -5,4 +5,28 @@ grant {
 
     permission java.net.SocketPermission "*", "connect,resolve";
     permission java.net.NetPermission "getProxySelector";
+
+    // needed because of problems in ClientConfiguration
+    // TODO: get these fixed in aws sdk
+    permission java.lang.RuntimePermission "accessDeclaredMembers";
```
skkosuri-amzn, line 11: Do we need `accessDeclaredMembers`? Could you please double check.

lezzago: It is required. The AWS SDK uses jackson databind, which needs `accessDeclaredMembers`.

kevinduke10:
Would love this merge!!

PR #360, opened 2021-03-09
cleanup: removed compile warnings, moved log msg, added missing copyr…

skkosuri-amzn:
*Issue #, if available:*
Cleanup

*Description of changes:*
Removed compile warnings in few Action classes.
Moved log msg from AlertingUtils
Added missing copyright in AlertingUtilsTests.

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
@@ -93,7 +93,7 @@ class TransportExecuteMonitorAction @Inject constructor(
                     }
                 })
             } else {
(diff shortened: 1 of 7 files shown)


lezzago (approved):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -93,7 +93,7 @@ class TransportExecuteMonitorAction @Inject constructor(
                     }
                 })
             } else {
-                val monitor = when (user == null || user?.name.isNullOrEmpty()) {
+                val monitor = when (user?.name.isNullOrEmpty()) {
```
qreshi, line 96: Interesting, I wasn't aware that user being null at the beginning would still call `isNullOrEmpty()` at the end of the call chain.

qreshi (approved):

PR #363, opened 2021-03-19
Refactor MonitorRunner

qreshi:
*Issue #, if available:*

*Description of changes:*
* Refactor MonitorRunner to use services
* Move existing logic into relevant services
* Existing code paths are already covered by MonitorRunner integration tests but will add unit tests for services later on

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

### DIFF
--- a/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertService.kt
@@ -1,0 +1,206 @@
+/*
+ *   Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
+ *   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *   express or implied. See the License for the specific language governing
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.alerting
+
+import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertError
+import com.amazon.opendistroforelasticsearch.alerting.alerts.AlertIndices
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.firstFailureOrNull
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.retry
+import com.amazon.opendistroforelasticsearch.alerting.elasticapi.suspendUntil
+import com.amazon.opendistroforelasticsearch.alerting.model.ActionExecutionResult
+import com.amazon.opendistroforelasticsearch.alerting.model.Alert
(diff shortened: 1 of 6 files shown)


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
```diff
@@ -66,80 +53,130 @@ import kotlinx.coroutines.launch
...
+
+        return this
+    }
+
+    // To be safe, call this last as it depends on a number of other components being registered beforehand (client, settings, etc.)
+    fun registerDestinationSettings(): MonitorRunner {
+        destinationSettings = loadDestinationSettings(settings)
+        destinationContextFactory = DestinationContextFactory(client, xContentRegistry, destinationSettings)
+        return this
     }
```
lezzago, line 177: Should we use Builders or just a constructor, so we can enforce these variables are not null?

lezzago (changes requested):

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
qreshi: I used a similar format to the runners in Index Management. The runner is a singleton now and should only be setup once in `AlertingPlugin.kt`. The class attributes are `lateinit`s which cannot be nullable. If `lateinit`s are accessed too early, they throw special exceptions. But since we register in the `createComponents()` method of `AlertingPlugin.kt` we should be fine.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/AlertService.kt
```diff
@@ -0,0 +1,206 @@
...
+            Alert(monitor = ctx.monitor, trigger = ctx.trigger, startTime = currentTime,
+                lastNotificationTime = currentTime, state = alertState, errorMessage = alertError?.message,
+                errorHistory = updatedHistory, actionExecutionResults = updatedActionExecutionResults,
+                schemaVersion = IndexUtils.alertIndexSchemaVersion)
+        }
+    }
+
+    suspend fun saveAlerts(alerts: List<Alert>, retryPolicy: BackoffPolicy) {
+        var requestsToRetry = alerts.flatMap { alert ->
+            // we don't want to set the version when saving alerts because the Runner has first priority when writing alerts.
```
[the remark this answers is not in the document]
qreshi: Updated

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/MonitorRunner.kt
lezzago: oh that is good to know, thanks

lezzago (approved):

PR #347, opened 2021-02-08
Add HTTP Input type for monitors (Inherited from PR #82)

tlfeng:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/alerting/issues/47

*Description of changes:*
This PR is inherited from https://github.com/opendistro-for-elasticsearch/alerting/pull/82,and here are the additional changes. I have also marked them in the code.
 - Re-base the code from opendistro-1.1 branch (ES 7.1.1) to opendistro-1.9 branch (ES 7.8.0)
 - Change the parameter name from `params` to `query_params` in the JSON string builder in funtion `HttpInput:toXContent()` to be able to reuse `URLInfo` component for the custom webhook URL in Kibana plugin
 - Change the URL path in the unit test `test monitor HttpInput with non JSON response ` due to the hidden index change
 - In `alerting/build.gradle`, add comments for the new configuration of the dependency versions, and remove `httpasyncclient` from dependency as it has been declared in `alerting-core` subproject

Minor changes: 
 - Change a function name from `validateFields` to `areFieldsValid` to because it returns a boolean value
 - Remove duplicate '+' from the error message for the above `validateFields()` validation function
 - Revert changing the orders of some `import` classes to reduce file differences of the PR

As the original PR has been merged (but reverted later on: https://github.com/opendistro-for-elasticsearch/alerting/pull/161, https://github.com/opendistro-for-elasticsearch/alerting/pull/162), I assume all the changes are reviewed, so I changed the code as less as possible.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #369, opened 2021-04-02
Local uri input monitor support

AWSHurneyt:
*Issue #, if available: [#347](https://github.com/opendistro-for-elasticsearch/alerting/pull/347)

*Description of changes:

1. Renamed assets from [#347](https://github.com/opendistro-for-elasticsearch/alerting/pull/347) from `HttpInput` to `LocalUriInput`
2. Implemented data class, LocalUriInput, to support creation of local cluster monitors using URI input.
2. Implemented logic for a map of supported API and their supported JSON payloads. Local cluster monitors for unsupported API will fail to execute.
3. Implemented logic to support creation of local cluster monitors for cluster health, and cluster stats using a LocalUriInput.

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.


alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
```diff
@@ -0,0 +1,66 @@
...
+         * Each Set<String> represents the supported JSON payload for the respective API.
+         */
+        private var supportedApiList = HashMap<String, Map<String, Set<String>>>()
+
+        /**
+         * Set to TRUE to enable the supportedApiList check. Set to FALSE to disable.
+         */
+        // TODO: Currently set to TRUE for testing purposes.
+        //  Should likely be set to FALSE by default.
+        private var supportedApiListEnabled = true
```
thalurur, line 27: just curious: why do we need this?

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
```diff
@@ -0,0 +1,66 @@
...
+            return supportedApiList[path] ?: throw IllegalArgumentException("API path not in supportedApiList: $path")
+        }
+
+        /**
+         * If [supportedApiListEnabled] is TRUE, calls [validatePath] to confirm whether the provided path
+         * is in supportedApiList. Will otherwise take no actions.
+         * @param localUriInput The [LocalUriInput] to validate.
+         * @return The path that was validated.
+         */
+        fun validateLocalUriInput(localUriInput: LocalUriInput): String {
```
thalurur, line 50: Do you think it makes sense to rename this to something like ``resolveToUriPath` as this method seems to be doing more than just validation, its also seems to be converting ``LocalUriInput`` to a path. 

Also do we really need to convert to a uri path? Looks the entity being used later on is transport action request do you think it makes sense to just build and return ActionRequest from this method instead, so callers don't need to worry about building the action request with all the correct attributes?

```
fun resolveToActionRequest(localUriInput: LocalUriInput): ActionRequest? {
  ....
} 
```

And caller can just call this method

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/SupportedApiSettingsExtensions.kt
```diff
@@ -0,0 +1,33 @@
...
+
+fun executeTransportAction(localUriInput: LocalUriInput, client: Client): ActionResponse {
+    val path = validateLocalUriInput(localUriInput)
+    if (path == SupportedApiSettings.CLUSTER_HEALTH_PATH) {
+        return client.admin().cluster().health(ClusterHealthRequest()).get()
+    }
+    if (path == SupportedApiSettings.CLUSTER_STATS_PATH) {
+        return client.admin().cluster().clusterStats(ClusterStatsRequest()).get()
+    }
+    throw IllegalArgumentException("Unsupported API: $path")
```
thalurur, line 22: minor:
you can use when statement instead 
```
when (path) {
   is SupportedApiSettings.CLUSTER_HEALTH_PATH -> return ..
   is SupportedApiSettings.CLUSTER_STATS_PATH -> return ...
   else -> throw IllegalArgumentException
}
```

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/SupportedApiSettingsExtensions.kt
```diff
@@ -0,0 +1,33 @@
...
+}
+
+fun ActionResponse.toMap(): Map<String, Any> {
+    if (this is ClusterHealthResponse) {
+        return this.convertToMap()
+    }
+    if (this is ClusterStatsResponse) {
+        return this.convertToMap()
+    }
+    throw IllegalArgumentException("Unsupported ActionResponse type: ${this.javaClass.name}")
```
thalurur, line 32: minor: when 
```
when (this) {
    is ClusterHealthResponse, is ClusterStatsResponse -> return this.convertToMap()
    else -> throw IllegalArgumentException
}
```

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
AWSHurneyt: This is essentially an on/off switch that I implemented at the beginning of the project. If a user did not want to check the API called by the monitor against the allow list, this value could be set to FALSE. That would allow users to create monitors for any API that this feature support. At the moment, it doesn't serve much of a purpose; but it could be more valuable as more API become supported.

We could remove this value if we feel it's extraneous.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
AWSHurneyt: That makes sense. I'll refactor this method to `resolveToActionRequest` and return the ActionRequest. Thank you for the feedback! I've also made the `validatePath` method public for use in the `Monitor` class.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/SupportedApiSettingsExtensions.kt
AWSHurneyt: Thank you for the tip! The syntax is a little different from your example due to some compiler warnings, but I refactored this to `when`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/util/SupportedApiSettingsExtensions.kt
AWSHurneyt: Thank you for the tip! The syntax is a little different from your example due to some compiler warnings, but I refactored this to `when`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
thalurur: If I understand this correctly this is something intended to be specified by the user and if this is set to false then they cannot use the LocalUrlInput feature.

I am not sure I understand how this is configurable by the end user at the moment. Do you think this can be defined as a cluster setting so user can turn it on off as it seems fit for them? or if this is not configurable at the moment suggest going with removing this as it serves no purpose and in future when needed to add the customization we can introduce the cluster setting

AWSHurneyt: If configured to false, users could still use the LocalUriInput feature; but the allow list wouldn't be checked for the requested API before creating and executing the monitor. At the moment, the allow list exposes both of the supported API; but as support for more API is added, a cluster owner could essentially configure this setting to false to allow access to any supported API, regardless of whether they are on the allow list.

Your comment about configuring via settings makes sense though. This will be easy enough to reimplement if there's a need in the future. 

Removed this value.

core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/model/LocalUriInput.kt
```diff
@@ -0,0 +1,169 @@
...
+            for (e in query_params.entries)
+                uriBuilder.addParameter(e.key, e.value)
+            uriBuilder.build()
+        } else {
+            URIBuilder(url).build()
+        }
+    }
+
+    /**
+     * Helper function to confirm at least [url], or [scheme]+[host]+[port]+[path]+[query_params] is defined.
```
thalurur, line 161: is it `either` or `alteast`? 

if its atleast then I'd think we need to check that 
either ``url`` and ``[scheme]+[host]+[port]+[path]+[query_params]`` is not empty and if they are both not empty then check they both are creating same URI?

If its either then do we need to check only one of them is non empty.


Looking at the code I am thinking either is what we are trying to do, but it seem to miss the validation for case that both are empty. We are only checking both are not not empty.

AWSHurneyt: Ah, thank you for pointing that out. Yes, we want **either** the `url` or the other fields to be provided, not both so we don't need to worry about a user providing two different URLs in one request. I've refactored this method to include that check.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
```diff
@@ -0,0 +1,91 @@
...
+         * @return The [ActionRequest] for the API associated with the provided [LocalUriInput].
+         */
+        fun resolveToActionRequest(localUriInput: LocalUriInput): ActionRequest {
+            val path = localUriInput.toConstructedUri().path
+            validatePath(path)
+            return when (path) {
+                CLUSTER_HEALTH_PATH -> ClusterHealthRequest()
+                CLUSTER_STATS_PATH -> ClusterStatsRequest()
+                else -> throw IllegalArgumentException("Unsupported API: $path")
+            }
```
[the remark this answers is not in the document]
AWSHurneyt: Ah, that's a really good point. Thank you for pointing that out! I removed the call to `validatePath`.

alerting/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/settings/SupportedApiSettings.kt
```diff
@@ -0,0 +1,91 @@
...
+         * }
+         */
+        private var supportedApiList = HashMap<String, Map<String, ArrayList<String>>>()
+
+        init {
+            val supportedJsonPayloads = SupportedApiSettings::class.java.getResource(RESOURCE_FILE)
+            @Suppress("UNCHECKED_CAST")
+            if (supportedJsonPayloads != null) supportedApiList =
+                XContentHelper.convertToMap(
+                    JsonXContent.jsonXContent, supportedJsonPayloads.readText(), false) as HashMap<String, Map<String, ArrayList<String>>>
```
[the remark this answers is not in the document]
AWSHurneyt: For P0, the `supportedApiList` is configured using the JSON; users will need to make changes to the resource file in order to update that list. As the plan was to only support `ClusterHealth` and `ClusterStats` for P0, an API to configure the list was not a high priority for P0.

